Issue #23, opened 2016-06-13
Add card description pop-up


Kazuo256:
Display card sprite and text side-by-side.

PR #25, opened 2016-06-14
implements #23 ticket - Display Card Description Popup

rafaeljpc:


scenes/main.gd
```diff
@@ -7,6 +7,8 @@ var player
 var map
 var done
 var next_sector
+var hand
+var displayPopup
```
Kazuo256: I know we haven't written any code conventions yet, but all other variables use snake case. Let's keep it that way for consistence until we decide otherwise =).

rafaeljpc: wow, sorry it's a java convention habbit I'll change it now

Kazuo256: Java ruins all of our lifes.

rafaeljpc: done

Issue #30, opened 2016-06-16
Upgrade cards


Kazuo256:
+ [x] A vanilla upgrade card
+ [x] A static effect upgrade card
+ [x] A trigger effect upgrade card

dread-uo:
What would static effect upgrade and trigger effect upgrade cards be?

Kazuo256:
Static effects are continuous effects, like "you get movement speed +2". They remain valid as long as the upgrade is installed.

Trigger effects are conditionally-timed effects, like "whenever you hit a monster, draw a card".

PR #35, opened 2016-06-20
#30 upgrade cards

rafaeljpc:


model/actor.gd
```diff
@@ -71,6 +74,18 @@ func step_time():
 	if can_draw():
 		draw_cooldown -= draw_rate
 
+func set_upgrade(upgrade):
+	if upgrades.size() == upgrade_slot:
+		upgrades.pop_front();
```
Kazuo256: The expected behavior from the design document was to not allow new upgrades if there are no slots available.

scenes/core/cards.gd
```diff
@@ -1,7 +1,11 @@
 
 extends Node
 
+const Card = preload("res://model/card_ref.gd")
+
 func get_card_id(card):
+	if not card extends Card:
+		assert(false)
```
Kazuo256: I believe this lines can be simplified to:
```gdscript
assert(card extends Card)
```

rafaeljpc:
Done

model/actor.gd
Kazuo256: What about this part? XD

rafaeljpc: I asked you at dircord about it....
I thought that is was possible to renew the upgrades until we make a better management upgrade system (or interface)

rafaeljpc: According with our conversation I updated the code

Issue #48, opened 2016-06-26
Add highlight effect for the tile in which a body is


OrenjiAkira:
Also maybe a less visible highlight in the adjacent walkable tiles, and test if these highlights work well for other bodies which are not the hero's, like monsters.

Kazuo256:
There is a highlight now, but it also shows under monsters.

OrenjiAkira:
I think that could be a feature so long as the monster highlight is different

Kazuo256:
Ok then I'll change this issue description and create a new one for the different color.

Issue #47, opened 2016-06-26
Make life bar depend on the body's sprite's height


OrenjiAkira:
As it is currently, the lifebar has a fixed height in relation to the tile in which its body inhabits. I believe it would be better if the lifebar was almost directly above the body's sprite, so its height in relation to the tile is variable, and would depend on the sprite's height.

Kazuo256:
The life bar are more aligned now, but it seems the sprites themselves are not (at least the slime isn't). Maybe it is related to how this feature was being implemented?

Kazuo256:
Anyone still working on this?

PR #51, opened 2016-06-27
issues #47 and #48

rafaeljpc:


components/bodyview.gd
```diff
@@ -17,6 +20,20 @@ func _ready():
...
+
+func unhighlight():
+	highlight = false
+	update_hl()
+
+func update_hl():
+	if highlight:
+		get_node("Highlight").show()
+	else:
+		get_node("Highlight").hide()
```
Kazuo256: Can't this method use the instance variable from line 10?

rafaeljpc: No,  it seems that variable is only ready after the unserialize process. Then during the load saved file it is null.

Kazuo256: I see. I'll merge it then.

PR #60, opened 2016-07-07
Controller Refactoring

rafaeljpc:


rafaeljpc:
@Kazuo256 done

Kazuo256:
I tested the branch and found a bug: when you cast the "teleport" card and try to move the target, the player character moves too.

rafaeljpc:
I'll see that O.o

rafaeljpc:
@Kazuo256 I pulled the branch to my Home-PC and it worked properly.

Kazuo256:
Yes, indeed. I tried again and it seemed to work now O.o

But then I found something else. When you are targeting something, pressing ESC should cancel the targeting mode, but in this branch it closes the game altogether.

rafaeljpc:
Okay I'll fix it

rafaeljpc:
The cursor_controller and main_controller were both handling the cancel event, I fixed it. But I think that the exit should be triggered by a popup before actually close the game. Maybe in another milestone

dread-uo:
About the popup, godot can handle it: http://docs.godotengine.org/en/latest/tutorials/engine/handling_quit_requests.html.

Kazuo256:
Okay, so now you only need to rebase this branch so we can merge it =).

PR #78, opened 2016-07-20
Laser beam card

dread-uo:


Kazuo256:
Is the PR ready? I'm gettin a `Identifier not found: CENTER` at `laser_beam.gd:51` here.

Kazuo256:
Also, if you want to, you can implement this card by adding a new type of OPTION to cards instead of using targets (the new OPTION would cast card in a direction instead of at some position). Don't worry it too much about it though.

dread-uo:
Sorry, forgot that last commit ¯\_(ツ)_/¯
I liked that OPTION alternative, it seems useful for future cards. I might implement it and refactor this later. Do you think the dynamic aoe feature is worth keeping?

Kazuo256:
> Do you think the dynamic aoe feature is worth keeping?

Absolutely =)

PR #140, opened 2017-08-11
@Kazuo256 is doing stuff

Kazuo256:
As soon as someone reviews this, feel free to merge. I'll keep on adding things otherwise XD.

This PR:

1. Unifies AI and player control of actors under the concept of "actor behavior"
2. Defines a movement enumeration, limiting movement to cardinal directions
3. Adds a placeholder graphical representation of actor bodies
4. Activates a simple debug GUI using the imgui lib
5. Moves the mapview class into a view folder
6. Draws checkered tiles
7. Makes the camera follow the player smoothly
8. Draw tiles with depth
9. Generate random holes in a larger map
10. Load body and actor specifications from database
11. Register body and actor ids in Route
12. Make actors carry only the body's id
13. Route register elements' subtypes as well


game/domain/actor.lua
```diff
@@ -5,13 +5,13 @@ local Actor = Class{
...
 
-function Actor:init(body, spec_name)
+function Actor:init(spec_name)
 
   GameElement.init(self, 'actor', spec_name)
 
   self.behavior = require('domain.behaviors.' .. self:getSpec 'behavior')
 
-  self.body = body
+  self.body_id = nil
```
OrenjiAkira: why nil and not false? just curious

Kazuo256: No real reason. Was worried with other things at the time.

game/domain/view/mapview.lua
```diff
@@ -36,24 +36,28 @@ function MapView:draw()
...
+        local x, y = j*TILE_W, i*TILE_H
         g.push()
-        g.translate(TILE_W/2, TILE_H/2)
-        g.scale(TILE_W, TILE_H)
-        g.setColor(200, 100, 100)
-        g.polygon('fill', 0.0, -0.8, -0.25, 0.0, 0.0, 0.2)
-        g.setColor(90, 140, 140)
-        g.polygon('fill', 0.0, -0.8, 0.25, 0.0, 0.0, 0.2)
+        g.translate(x, y)
+        g.setColor(tile)
```
OrenjiAkira: I'm not sure about this representation of tiles. Obviously it will change later on.

Kazuo256: Yup.

OrenjiAkira:
It looks good to me in general. There are some things that are temporary, but overall they make the game nicely debuggable for now, which is what's important at the moment. The only thing that sticks out to me is that we need to think about a more defined common map representation system.

PR #157, opened 2017-08-18
Map generation modules

OrenjiAkira:
Adds transformation modules for making maps in `game/domain/transformers/`.

An example of usage is available in `game/tests/map1.lua`. It is reasonably data-driven-friendly.
You can also run bulk tests running `game/tests/bulkmap.lua`. If you pass a number to that file, it will be used as a seed, otherwise it will run 256 different random seeds for testing.


game/domain/transformers/helpers/random.lua
```diff
@@ -0,0 +1,33 @@
+
```
Kazuo256: This file should be at `game/common`

OrenjiAkira:
Requested changes have been done.

PR #158, opened 2017-08-18
Ai control + action format

### LINKED ISSUE #154: AI can't pick target
Because we always delegate to the player

### LINKED ISSUE #160: Action format
We need a way to represent actions. Initial suggestion:

Actions have three lists: parameters, values, and effects.

1. Parameters are used to receive input from actors (their behaviors actually)
2. Intermediate values computed for the effects
3. The actual effects of the action

Example:
```lua
action = {
  params = {
    { "choose_target", {5, "BODY"}, "target" } -- body-only target within range 5
  },
  values = {
    { "get_attribute", {"ATH"}, "attr" }, -- gets actor attribute
    { "add", {"attr", 3}, "amount" }, -- sums results from value "attr" with constant (3)
  },
  effects = {
    { "deal_damage", { "val:amount", "par:target" } } -- deals ATH+3 damage to target body
  }
}
```

Kazuo256:
1. Actions are now structs of parameters, values and effects (as per #160)
2. Behaviors choose action by name (like MOVE or PRIMARY)
3. Signal handlers in play state are coroutines, and handle multiple action parameters
4. AI is now able to pick targets. In fact, it may attack you when you least expect!

Minor refactorations:
1. Refactorate an `IDGenerator` out of `Route`
2. Removed unused local variables from game states

Close #154
Close #160


OrenjiAkira:
I noticed the AI does not discriminate player and other monsters. That sounds cool at first, but is there a way for them to actually discriminate?

Kazuo256:
There will likely be eventually. It was not intentional for AIs to not discriminate, it's just that target discrimination was not part of the objetive of this PR.

PR #164, opened 2017-08-20
Map/Sector construction now uses previously implemented generation algorythms

### LINKED ISSUE #161: Load map tiles from generator
Currently, `domain.Map` fills its own tiles randomly. Now that we have a proper generator, it should only initialize its internal tile structure in the contrusctor then later have its tiles set. Which means we need to:

1. Remove the random tile filling code from `domain.Map:init()`
2. Add a tile setter method to `domain.Map`
3. Genrate a random map in `gamestates.play:enter()`
4. Feed the generated tiles into `_current_map`

OrenjiAkira:
Related to: #161 

All I did was:

1. Add a new `json` file to the `database` directory. It contains the parametres for constructing a sector.
2. Change `Map:init()` to receive the database entry name instead of width and height. It now runs the generation and transformer modules in `domain/transformers/`, and creates the sector based on the generated `MapGrid` object.

### DIFF
--- a/game/database/sector1.json
+++ b/game/database/sector1.json
@@ -1,0 +1,40 @@
+{
+  "general": {
+    "width": 32,
+    "height": 32,
+    "mw": 3,
+    "mh": 3,
+  },
+  "transformers": [
+    {
+      "name": "rooms",
+      "params": {
+        "minw": 3,
+        "minh": 3,
+        "maxw": 7,
+        "maxh": 7,
+        "count": 12,
+        "tries": 256,
+      }
+    },
+    {
+      "name": "maze",
+      "params": {
+        "double": false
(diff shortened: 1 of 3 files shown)


Kazuo256:
Looks good. The only issue I see is that you used the whole sector domain for the spec of a single map archetype. With bodies and actors, each domain has a list of specs in a single file.

OrenjiAkira:
I am not sure I understand. In `gamestates/play.lua` you pass which sector-specs you want to use. Is that not enough for now?

Kazuo256:
Well, yes. We can just open an issue for future reference then. In fact, loading from the database is already a big bonus from what the issue asked. I'll merge this in a few.

PR #166, opened 2017-08-20
Make sector db one file with many entries

### LINKED ISSUE #165: Store sector specs in database the same way as bodies and actors
Bodies and actors each have a single JSON file in the database containing the entires for every spec in the corresponding domain. On the other hand, sector specs are currently stored on separate, individual files. Since the `database` module was designed for the former usage, it would be best if sectors followed the same storage structure.

OrenjiAkira:
Closes #165


game/domain/map.lua
```diff
@@ -14,7 +14,9 @@ function Map:init(sector_name)
 
   ELEMENT.init(self)
 
-  local specs = DB.loadDomain(sector_name)
+  local specs = SECTORS[sector_name]
```
Kazuo256: If possible, use `DB.loadSpec(domain_name, spec_name)` instead.

OrenjiAkira: Alrighty.
My reasoning for this was that well, if we are using only the sector-section of the DB _and_ it's only one file, we might as well save that entry itself in a "const" in the Sector class.

I suppose calling`DB` explicitly makes it clearer to other people what is happening.

Kazuo256: > I suppose calling `DB` explicitly makes it clearer to other people what is happening.

That, and it keeps things decoupled: the code using `DB` doesn't need to know how specs are actually stored inside domains.

PR #177, opened 2017-08-26
Basic profile module (does not save routes yet)

OrenjiAkira:
This adds a profile module in `infra/profile.lua` and a runflags module in `infra/runflags.lua`.

The `PROFILE` module takes care of saving the game's metadata and its savefiles. Its `save` and `init` functions are for loading/creating the metadata, currently saved in a json file in your user's save directory. The `saveRoute` and `loadRoute` functions are for saving/loading individual routes. There is also a `newRoute` function, which currently only generates a route id.

We have no savefiles yet, though. This PR is just for adding a save/load system.

Also, for debugging purposes, I added runflags. You can run the game using `love` directly and pass parametres like `-clear` or `--clear` and they are stored in all caps in `RUNFLAGS.CLEAR`. Since we have a visual dev mode editor already, I thought it might be a tad... rustic. But it makes deleting savefiles and profiles easy, and this is why I added them with this PR. This could be used to allow dev mode too, maybe. I don't know. Check it out.

### DIFF
--- a/game/gamestates/start_menu.lua
+++ b/game/gamestates/start_menu.lua
@@ -52,8 +52,8 @@ function state:update (dt)
   if MENU.begin(_menu_context, 80*4, _height / 2, 3, 160) then
     if _menu_context == "START_MENU" then
       if MENU.item("New route") then
-        local route_id = PROFILE.newRoute()
-        SWITCHER.switch(GS.PLAY, route_id)
+        local route_data = PROFILE.newRoute()
+        SWITCHER.switch(GS.PLAY, route_data)
       end
       if MENU.item("Load route") then
         _menu_context = "LOAD_LIST"
@@ -64,8 +64,8 @@ function state:update (dt)
     elseif _menu_context == "LOAD_LIST" then
       local savelist = PROFILE.getSaveList()
       if #savelist > 0 then
-        for _,route_id in ipairs(savelist) do
-          if MENU.item(route_id) then
+        for route_id, route_header in pairs(savelist) do
+          if MENU.item(route_header.charname) then
             SWITCHER.switch(GS.PLAY, PROFILE.loadRoute(route_id))
           end
         end
--- a/game/gamestates/start_menu.lua
+++ b/game/gamestates/start_menu.lua
@@ -2,6 +2,7 @@
 local MENU = require 'infra.menu'
 local INPUT = require 'infra.input'
 local CONTROLS = require 'infra.control'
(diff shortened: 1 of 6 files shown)


game/gamestates/start_menu.lua
```diff
@@ -46,18 +49,38 @@ end
...
+      end
+      if MENU.item("Quit") then
+        love.event.quit()
+      end
+    elseif _menu_context == "LOAD_LIST" then
+      local savelist = PROFILE.getSaveList()
+      if #savelist > 0 then
+        for _,route_id in ipairs(savelist) do
+          if MENU.item(route_id) then
+            SWITCHER.switch(GS.PLAY, PROFILE.loadRoute(route_id))
```
Kazuo256: On line 56 you passed the route ID to the PLAY gamestate, but here you pass the actual route data. Which one is the intended usage?

OrenjiAkira: I have no idea. We need to determine what builds the save. It might be the route, the profile, or it might be something else entirely. I guess the `PROFILE` can fetch the route data from the savefile, but can it, or should it, generate route data when creating a new route?

Kazuo256: I guess it's probably better to have a separate class `RouteGenerator` or equivalent (Maker? Builder?). `PROFILE` uses requests fresh new route data from it and passes to `PLAY` gamestate. Something like that?

Kazuo256: Though this is likely not necessary for this PR.

OrenjiAkira: That sounds good, I think. For now I guess it can be a simple table generator? I will can add it to the PR

Kazuo256: OK

OrenjiAkira:
I did what you suggested and created a new module `ROUTEBUILDER` in `game/infra`. It generates an semi-empty sample savedata table with what fields I could remember that a route would need. `PROFILE` module now uses that for creating new files, and thus `START_MENU` now passes consistently the `route_data` to the `PLAY` gamestate.

I tweaked some other stuff too. I added a version field to the `profile` savedata, and I changed the `save_list` to use the `route_id` as the field name, and the content will be the route's header (I suggest putting in it info like `character name` and `last save date` maybe; all for displaying purposes in the `START_MENU`). I also improved the `RUNFLAGS` module to accept values to the flags (only string and numbers).

PR #224, opened 2017-09-02
Store only type in tiles

Kazuo256:
That is, make tiles no longer hold their color.

This makes all floor tiles look the same (no more checkered pattern!), but simplifies how saves are stored and loaded. Besides, `SectorGrid:from(tiles)` seems to be no longer necessary now. Should I remove it?

I am actually not 100% sure of this PR.

### DIFF
--- a/game/domain/sector.lua
+++ b/game/domain/sector.lua
@@ -24,6 +24,8 @@ local function _initBodies(w, h)
       t[i][j] = false
     end
   end
+  -- A special tile where we can always remove things from...
+  -- Because nothing is ever there!
   t[0] = { [0]=false }
   return t
 end
@@ -36,14 +38,10 @@ function Sector:init(spec_name)
   self.h = 1
 
   self.tiles = {{ false }}
-  self.bodies = {}
+  self.bodies = _initBodies(1,1)
   self.actors = {}
   self.exits = {}
   self.actors_queue = {}
-
-  -- A special tile where we can always remove things from...
-  -- Because nothing is ever there!
-  self.bodies[0] = { [0] = false }
 
   self.turnLoop = coroutine.create(_turnLoop)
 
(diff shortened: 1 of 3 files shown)


OrenjiAkira:
I think this is a good PR. It's halfway to using actual tiles. I guess you can remove the `SectorGrid:from` method since we now save the tiles the same way we use them. The only exception is the `NAUGHT` tile. One could argue that when saving `false` occupies 5 chars meanwhile `' '` occupies three, so maybe empty tiles could be saved as is without converting them to `false`. But ehh, we'll see about it when we actually need to change this code.

OrenjiAkira:
I'm leaving removing the `SectorGrid:from` method to your opinion, so you can merge it as is if you like.

Kazuo256:
> The only exception is the NAUGHT tile. One could argue that when saving false occupies 5 chars meanwhile `' '` occupies three

The PR actually saves tiles as `{ type = ' ' }`. Is that still ok?

OrenjiAkira:
That's actually better.

game/domain/sector.lua
```diff
@@ -134,16 +131,9 @@ function Sector:makeTiles(grid)
...
-      if grid.get(j, i) == SCHEMATICS.FLOOR then
-        if (i+j) % 2 == 0 then
-          tile = { unpack(COLORS.FLOOR1) }
-        else
-          tile = { unpack(COLORS.FLOOR2) }
-        end
-      elseif grid.get(j, i) == SCHEMATICS.EXIT then
-        tile = { unpack(COLORS.EXIT) }
+      if tile_type and tile_type ~= SCHEMATICS.NAUGHT then
+        tile = { type = tile_type }
```
OrenjiAkira: Are you sure it saves as `' '`? Here it transforms into false, and then when you save it will stay false.

Kazuo256: Oh, right. I meant to say tiles didn't use just 3 characters, but indeed the `NAUGHT` tile is still saved as `false`. Want me to treat it the same as the others?

OrenjiAkira: I think it's better to treat it the same if we're not going to have any particular use for the boolean value. The sectorview code, however, does become shorter with the false value. At this point in the development I don't think there is a difference.

OrenjiAkira: I just thought that the char value is actually "hasheable" meanwhile false isn't. So there's that.

Kazuo256: I tried to change it to follow the other types of tiles, but realized there's actually a number of places in the code that assume false tiles mean empty tiles. I'll open an issue and leave it for later.

Issue #227, opened 2017-09-04
Enhance exit transformers


OrenjiAkira:
Currently a possible exit is any floor tile that has four directly adjacent tiles that are also floors. Having eight would already be a great improvement.

OrenjiAkira:
closed via #234

PR #234, opened 2017-09-05
Make exits placement have better thresholds

OrenjiAkira:
Makes exits only appear 5 tiles appart, in the middle of an area of 5x5 floor tiles. A little less performance, but it's still linear. I think. Not sure if these threshold values should be parametres.

Also fixes DB usage in the sector01 generation test. I cannot believe it was even working without that fix, but it was, albeit with unexpected results.

Should help with #227 
Might conflict with #235

### DIFF
--- a/game/domain/transformers/exits.lua
+++ b/game/domain/transformers/exits.lua
@@ -66,7 +66,7 @@ function transformer.process(sectorinfo, params)
   end
 
   -- get a number of random possible exits from that list
-  (function()
+  local function chooseExits()
     local N = #exits_specs -- max number of exits
     for edx = 1, N do
       local i, j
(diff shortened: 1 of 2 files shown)


game/domain/transformers/exits.lua
```diff
@@ -16,47 +16,91 @@ transformer.schema = {
...
+      if _isPossibleExit(sectorgrid, x, y) then
+        table.insert(possible_exits, {y, x})
+      end
     end
   end
 
-  sectorinfo.exits = getRandomExits(getPossibleExits())
-  setExits(sectorinfo.exits)
+  -- get a number of random possible exits from that list
+  (function()
```
Kazuo256: You can change this for a `do ... end` block.

PR #241, opened 2017-09-06
Add attributes to actors

OrenjiAkira:
pretty straightfoward

### DIFF
--- a/game/database/actor.json
+++ b/game/database/actor.json
@@ -1,8 +1,14 @@
 {
   "player":{
-    "behavior":"player"
+    "arc":0,
+    "behavior":"player",
+    "ath":0,
+    "mec":0
   },
   "dumb":{
-    "behavior":"random_walk"
+    "arc":0,
+    "behavior":"random_walk",
+    "ath":0,
+    "mec":0
   }
 }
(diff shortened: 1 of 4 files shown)


Kazuo256:
There is also the `get_attribute` operator to fix.

OrenjiAkira:
what?

Kazuo256:
In `game/domain/operators/get_attribute.lua`. There's a `FIXME` pointing where the attributes should be used.

OrenjiAkira:
ok, but what is it supposed to do? is it used anywhere so i can see it?

Kazuo256:
It's used in the `SHOOT` action if I'm not mistaken.

OrenjiAkira:
ok, i have no idea of what is going on with that code

Kazuo256:
OK, maybe the `take_damage` effect better illustrates how to write the actual code. Did you also look at the `SHOOT` action through devmode?

The idea is that the fields specified in the operator's schema are available in the `params` argument of the `OP.process()` function. In the case of `get_attribute` it determines which attribute to get. All you have to do is return the corresponding attribute value from the actor the function receives.

OrenjiAkira:
like this?

Kazuo256:
Pretty much. Maybe we'll eventually add something to make sure a weird getter isn't called.

PR #245, opened 2017-09-07
Widget selector

Kazuo256:
Added a simple menu for choosing widgets.

Controls in this PR become as follows:
+ ACTION_1 (R button) still launches the primary action of the actor
+ ACTION_2 (E button) will keep the widget menu open while held down
+ ACTION_3 (W button) and ACTION_4 (Q button) are unused for now

While the widget menu is visible, you can select a widget with the directionals, and confirm with the CONFIRM input (F button).

Try it out and tell me how it feels!

### DIFF
--- a/game/domain/definitions/dir.lua
+++ b/game/domain/definitions/dir.lua
@@ -1,6 +1,6 @@
 
 return {
-  'left', 'right', 'down', 'up',
+  'right', 'up', 'left', 'down',
   left = {0,-1},
   right = {0,1},
   down = {1,0},
--- a/game/domain/view/widgetview.lua
+++ b/game/domain/view/widgetview.lua
@@ -21,8 +21,15 @@ function WidgetView:init(route)
 
   self.route = route
   self.invisible = true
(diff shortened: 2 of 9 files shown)


OrenjiAkira:
Have you tried using a widget on top of an exit?

I'm feeling a little weird about pressing confirm when selecting a widget, but I think that's probably the lack of sound feedback.

rilifon:
I'm still a bit confused on what every button does and how my hand should be on the keyboard haha

rilifon:
We should maybe map some button to the right side of the keyboard since everything is made for left-hand only. Dividing the buttons could better divide and simplify for the player what interactions he can make

Kazuo256:
I think you can change the input layout in your game profile already =D.

OrenjiAkira:
```


    Have you tried using a widget on top of an exit?

Yes, seems to work fine here.
```
No, you go to the other sector when you do it. :(

Kazuo256:
Should be fixed now.

PR #246, opened 2017-09-07
Add profile/savedata compression with external lualzw lib

OrenjiAkira:


Makefile
```diff
@@ -76,6 +79,16 @@ $(CPML_REPO):
 $(DKJSON_LIB):
 	wget -O $(DKJSON_LIB) -- http://dkolf.de/src/dkjson-lua.fsl/raw/dkjson.lua?name=16cbc26080996d9da827df42cb0844a25518eeb3
 
+## LUALZW
+
+$(LUALZW_LIB): $(LUALZW_REPO)
+	mkdir $(LUALZW_LIB)
+	cp -r $(LUALZW_REPO)/* $(LUALZW_LIB)
+	echo "return require 'lualzw.lualzw'" | cat >$(LUALZW_LIB)/init.lua
```
Kazuo256: Or just
```bash
echo "return require 'lualzw.lualzw'" > $(LUALZW_LIB)/init.lua
```

Kazuo256:
Can we have the option NOT to compress, for debugging purposes?

Makefile
OrenjiAkira: oh, right

OrenjiAkira:
do we use compressing as non-default for now then?

rilifon:
Something I thought today: should we keep using the makefile to compile this external libs we don't own eveytime? When we make a release but they update their lib and breaks our game, shouldn't we have a stable version to rely on?

Kazuo256:
We just have to make the deployed version of our game carry the approriate versions of the dependencies. A `make deploy` command for that will also help.

PR #247, opened 2017-09-08
Route exit action

### LINKED ISSUE #204: Make an action for changing sectors

### LINKED ISSUE #230: Move sector exit code to Route
Specifically, move most of the code from `_exitSector()` in `gamestates/play.lua` to `Route`.

Kazuo256:
Make a `CHANGE_SECTOR` action and use it when interacting with exits.

Close #230 
Close #204

### DIFF
--- a/game/main.lua
+++ b/game/main.lua
@@ -1,4 +1,7 @@
 
+-- LUX portability
+require 'lux.portable'
+
 -- CPML
 cpml      = require 'cpml'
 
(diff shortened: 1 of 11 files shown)


game/main.lua
```diff
@@ -1,4 +1,7 @@
 
+-- LUX portability
+require 'lux.portable'
```
OrenjiAkira: I was gonna suggest we did the things this file does eventually. Glad it's done now

PR #248, opened 2017-09-09
Add tiles drawn from spritebatch

OrenjiAkira:
No one asked me to do this, but I thought that eventually we will have to worry about dealing with sprites, so why not start small.

1. Made a very simple tile spritesheet. Made it draw instead of rectangles.
2. Minor tweaking in draw method of sectorview.

This is WIP because I am awating feedback on whether this code should stay provisory the way it is or not.


game/domain/view/sectorview.lua
```diff
@@ -7,10 +7,17 @@ local TILE_H = 80
...
+local TEXTURE = love.graphics.newImage("assets/imgs/tiles.png")
+TEXTURE:setFilter("nearest", "nearest")
 local TILE_COLORS = {
-  [SCHEMATICS.FLOOR] = COLORS.FLOOR1,
-  [SCHEMATICS.EXIT] = COLORS.EXIT,
+  [SCHEMATICS.FLOOR] = love.graphics.newQuad(0, 0, TILE_W, TILE_H, TEXTURE:getDimensions()),
+  [SCHEMATICS.EXIT] = love.graphics.newQuad(80, 0, TILE_W, TILE_H, TEXTURE:getDimensions()),
+  shade = love.graphics.newQuad(160, 0, TILE_W, TILE_H, TEXTURE:getDimensions()),
 }
+local TILES = love.graphics.newSpriteBatch(TEXTURE, 512, "stream")
```
Kazuo256: I think my only complain is this part where you call functions from `love.graphics` "globally". With or without using the DB, it's generally better if this kind of code is in some sort of init function (in this case, the `SectorView`'s constructor should do).

OrenjiAkira: reasonable enough, on it

PR #251, opened 2017-09-09
Card play design [take 1]

Kazuo256:
This PR experiments with one of the most recent ideas I had for card and hand dynamics in Backdoor.

The cardplay mechanic follows closely the one seen in deck-building card games such as Star Realms (I have it on my cell phone in case any of you want to check it out):

1. You have a deck which you are constantly adding cards to
2. On your "turn", you draw your whole hand, then play until you run out of cards

Adapting to Backdoor's context, it essentially means you can't draw another hand until you've finished the previous one. To avoid a bad hand locking the player, we just need to add an option to throw away cards (with possibly a compensation of sorts or not). This PR does not include this discard option (I can add it if you want to test it out).

**I chose the EXTRA (A button) input for the new "draw hand" action.**

Other things to consider:

1. For now, there are only three cards (damage, heal, and draw), so there isn't much to test
2. Ideally, you'd have a number of decks in the game, and you choose which to draw your hand from each time
3. There would also be a resource (card points?) you need to spend in order to draw a hand

### DIFF
--- a/game/database/action.json
+++ b/game/database/action.json
@@ -17,7 +17,7 @@
     "operators":[],
     "effects":[{
         "typename":"draw_card",
-        "amount":2
+        "amount":1
       }]
   },
   "CHANGE_SECTOR":{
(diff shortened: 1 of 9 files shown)


rilifon:
Uhh I really really liked the ideia! Its my favorite one so far. I'll try out here how it plays in-game and comment later

PR #262, opened 2017-09-13
Add renaming option for domain specs

### LINKED ISSUE #209: Rename stuff on devmode
Make user able to rename stuff and store the new names for stuff (actions specs, body specs, etc)

Kazuo256:
Close #209


OrenjiAkira:
Crashes when renaming currently used specs.

Kazuo256:
Yeah, there is not much to do here. If you rename a spec, every object in the game using that spec will crash.

PR #278, opened 2017-09-15
Buffers

Kazuo256:
Add buffers from which actors can draw their cards from.

There are 3 initial buffers by default, each containing 4 copies each of "bolt", "cure" and "think".

Also, cards in hand can now be recalled or consumed (provisory terms).


.gitignore
```diff
@@ -3,4 +3,8 @@
 .*
 !.git*
 /imgui*
+/game/cpml*
+/game/dkjson.lua
+/game/lux*
+/game/steaming*
```
OrenjiAkira: who is responsible for this???? this is 10000% unnecessary, and already resolved by putting everything in the `game/lib/` dir, stop this immediately

Kazuo256: @rilifon probably did this after noticing the files appered as untracked (instead of ignored) by git. This happened because we moved the files before and he probably didn't run `make clean`, thus  leaving the obsolete files in their old locatiom in additiom to the correct one.

All in all, I think it's very understandable that he did this. Although we do have to remove it. Will do tomorrow.

game/draw.lua
```diff
@@ -24,6 +24,8 @@ function draw.allTables()
 
     CAM:detach() --Stop tracking camera
 
+    DrawTable(DRAW_TABLE.HUDl)
```
OrenjiAkira: was it really necessary to create a whole new layer for this? the regular hud layer should have been fine, no?

game/setup.lua
```diff
@@ -41,7 +42,9 @@ function setup.config()
     CAM = Camera(O_WIN_W/2, O_WIN_H/2) --Set camera position to center of screen
 
     --IMAGES--
+    local path = "assets/imgs/"
     IMG = { --Table containing all the images
+      CARD_BACK_1 = love.graphics.newImage(path.."cards/card_back_1.png")
```
OrenjiAkira: i'll let this slide on account that i also made a hardcoded image asset loading on the `sectorview`, but we should use the DB for loading textures from image assets

game/draw.lua
Kazuo256: The problem is that there is no guaranteed drawing order between elements in the same layer. In this case, @rilifon wanted to add a transluscent background to the HUD and, in order to make sure it is drawn at the right time, the only way is to rely on the layer mechanism.

However, searching the code for "L2" shows that this layer is still unused. Should we use it instead? Won't we need it for anything else?

OrenjiAkira: ah okay, but we should rename them so they make sense. like, why is HUD1 drawn before HUD? we can do this in a different PR tho

rilifon: its not HUD1, its HUDl for lower. I made a coment about it on the setup.lua file.

@Kazuo256 comments explains my line of thought

.gitignore
rilifon: @Kazuo256 comments explains my line of thought

PR #288, opened 2017-09-19
Add new transformer to fill weird lone empty tiles

OrenjiAkira:
This new transformer fills in empty tiles when there is only one adjacent empty tile to it. It looks very different than before, but I think it's good. Check it out and test it before accepting.


Kazuo256:
Da fills

PR #304, opened 2017-09-23
Add aggro behavior to monsters

OrenjiAkira:
They now chase your ass down and kill you mercilessly.
This lags if there are lots of monsters chasing you down a narrow one-way path. The workaround for now is limiting their 'vision', and also them killing you before they can group significantly around you.

EDIT: Oh, right. This uses a different distance calculation than we currently to check when a monster is within range. Might need some discussing on whether we want to walk/attack in diagonals.

### DIFF
--- a/game/domain/behaviors/aggro.lua
+++ b/game/domain/behaviors/aggro.lua
@@ -82,7 +82,7 @@ return function (actor, sector)
   if dist == 1 then
     -- attack if close!
     return 'PRIMARY', { target = {target:getPos()} }
-  elseif dist <= 8 then
+  else
     -- chase if far away!
     local start = os.clock()
     local pos = _findPath({i,j}, {target:getPos()}, sector)
--- a/game/domain/behaviors/aggro.lua
+++ b/game/domain/behaviors/aggro.lua
@@ -5,6 +5,8 @@ local TILE = require 'common.tile'
 local Heap = require 'common.heap'
 
 local abs = math.abs
+local VISION = 20 -- high value to test heap efficiency
+local DEDICATION = math.floor(VISION * math.sqrt(2))
 
 local function _hash(pos)
   if not pos then return "none" end
@@ -23,12 +25,14 @@ local function _findPath(start, goal, sector)
   local cost_so_far = {}
   local path = {}
   local found = false
+  local count = 0
 
   frontier:add(start, 0)
   came_from[_hash(start)] = true
   cost_so_far[_hash(start)] = 0
 
   while not frontier:isEmpty() do
+    count = count + 1
     local current, rank = frontier:getNext()
 
     -- if you found your goal, quit loop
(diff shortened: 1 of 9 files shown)


game/domain/behaviors/aggro.lua
```diff
@@ -0,0 +1,95 @@
...
+      found = true
+      goal = current
+      break
+    end
+
+    -- look at neighbors
+    for n = 1, 4 do
+      local i, j = unpack(current)
+      local ti, tj = unpack(goal)
+      local di, dj = unpack(DIR[DIR[n]])
```
Kazuo256: If you prefer, you can iterate directly over `DIR` instead:
```lua
for _,dir in ipairs(DIR) do
  local di, dj = DIR[dir]
end
```

OrenjiAkira:
WIP or no WIP, I need feedback on those problems I mentioned.

Kazuo256:
I think limiting the vision should make a good difference. If it doesn't, we can either use a relaxed A-star or retort to "dumber" algorithms (as in, walk straight in the target's direction).

The relaxed A-star is very simple, just multiply the weight of the distance heuristic by 2. It stops finding the best paths, but is still better than "dumb" and considerably faster.

As for the walking and attacking diagonally, I'm not 100% sure yet but the more I think about it the more it seems to work better overall.

OrenjiAkira:
That cannot be that easy, I swear to god...

OrenjiAkira:
Ok, I saw literally no difference in multiplying the dist heuristic by two

Kazuo256:
[Like this?](https://en.wikipedia.org/wiki/A*_search_algorithm#Bounded_relaxation)

Kazuo256:
You saw no difference in efficiency?

OrenjiAkira:
I will test it more properly to see, but from playing I still felt around the same ammount of lag

OrenjiAkira:
according to my data, the tests yielded similar lag and time usage.
i'm using code nearly identical to this: https://www.redblobgames.com/pathfinding/a-star/implementation.html#python-astar

Kazuo256:
> multiplying the heuristic would just make priorities have bigger values

It changes how tiles are prioritized in the path, by making tiles "closer" to the target "better", even if they are further away from the starting point.

But if ain't improving speed, then that's that. As far as I can see though, the current implementation still allows actors to "see" the whole map, right?

OrenjiAkira:
do you mean "seeing" by:

1. Seeing a player within a certain distance and then choose to chase it (thus running the algorythm)

or

2. When running the algorythm, not looking at anywhere too far away?

Kazuo256:
> 2. When running the algorythm, not looking at anywhere too far away?

This.

OrenjiAkira:
Then, yes, it does look at the entire map for a possible solution until it is clear that no solution is available.

How would you suggest going around that?

OrenjiAkira:
Should be enough for a merge, but review properly first

Issue #305, opened 2017-09-25
Diagonal movement


Kazuo256:
It seems necessary.

Possible inputs:
+ Numpad
+ HJKL + YUBN
+ WASD + QEZC

PR #311, opened 2017-09-27
Add player diagonal movement

OrenjiAkira:
No support for joystick/gamepad yet


Kazuo256:
From #305, you must also support numpad directionals.

OrenjiAkira:
You do know that without the "numlock" thing, the numbers become pageup, pagedown, home and end, right?
But I added the numbers anyway, but now you can use the actual numbers too, I guess

PR #319, opened 2017-09-28
Move animation

### LINKED ISSUE #134: Implement move animations
Slide actor sprite when they move on their turns.

Kazuo256:
Movement effect yields turn coroutine requesting movement animation in a per-body basis.

We'll need to refactor this for more general animations soon, but I think it's best to merge this for the moment. Of cource, please test it before accepting the PR =).

Close #134

### DIFF
--- a/game/view/sector.lua
+++ b/game/view/sector.lua
@@ -80,7 +80,7 @@ function SectorView:addVFX(extra)
     local i0, j0 = body:getPos()
     local offset = {i - i0, j - j0}
     self.vfx.offset[body] = offset
-    self:addTimer(nil, MAIN_TIMER, "tween", 0.10, offset, {0, 0}, "in-out-quad",
+    self:addTimer(nil, MAIN_TIMER, "tween", 0.05, offset, {0, 0}, "in-out-quad",
                   function() self.vfx.offset[body] = nil end)
   end
 end
--- a/game/view/sector.lua
+++ b/game/view/sector.lua
@@ -55,6 +55,7 @@ function SectorView:init(route)
 
   self.target = nil
   self.cursor = nil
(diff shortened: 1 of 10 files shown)


game/view/sector.lua
```diff
@@ -63,10 +66,25 @@ function SectorView:init(route)
...
   self.target = target
 end
 
+function SectorView:addVFX(extra)
+  if extra.type == 'body_moved' then
+    local body, i, j = extra.body, unpack(extra.origin)
+    local i0, j0 = body:getPos()
+    local offset = {i - i0, j - j0}
+    self.vfx.offset[body] = offset
+    self:addTimer(nil, MAIN_TIMER, "tween", 0.10, offset, {0, 0}, "in-out-quad",
```
OrenjiAkira: This is kinda slow, esp when each body moves at a time. I tested with 0.05 and it looked fine. The faster the better. The flickering of the actor hud is bothersome but we can hide it by default so it doesn't keep flickering all the time.

OrenjiAkira:
It still bothers me a little that the bodies move one at a time. Is this something we will keep or will we try to do something about it in the future?

Kazuo256:
Changed animation time to 0.05. I had actually set it that way, but changed back to test something and forgot to revert it >.<

Also
> It still bothers me a little that the bodies move one at a time. Is this something we will keep or will we try to do something about it in the future?

Doing one animation at a time gives support for consecutive move animations out-of-the-box, which is something @rilifon thought was important to have. It *is* possible to do with simultaneous animations, but that would require changing alsmot everything about how we draw bodies. Basically, we wouldn't be able to use their "current position" to place them when drawing.

PR #323, opened 2017-09-30
Atomize database

OrenjiAkira:
This PR rearranges all of the database to use smaller files that contain only a specified spec. It flexibilizes the database cache too, though it works best if you use it with consistent levels of depth.

In the future, we should  do two things:
+ Allow atomic saves (save one item only, instead of all loaded entries)
+ Use the dkjson keyorder feature through schemas. This would make diffs much better.


OrenjiAkira:
i suggest ignoring e987c24 as it is really big and envolves splitting the json files

game/database/init.lua
```diff
@@ -46,24 +35,94 @@ function _subschemaFor(base, branch)
...
+  local file = assert(io.open(_fullpath(relpath), 'w'))
+  local data = json.encode(rawdata, {indent = true})
+  assert(file:write(data))
+  return file:close()
+end
+
+local function _save(cache, basepath)
+  -- check if file is marked to be deleted
+  if cache == DEFS.DELETE then
+    return assert()
```
Kazuo256: What.

game/database/init.lua
```diff
@@ -79,39 +138,80 @@ function DB.schemaFor(domain_name)
...
+    end
+
+    -- if json file
+    local filepath = path..".json"
+    if fs.exists(filepath) then
+      obj = _loadFile(filepath)
+      DB.initSpec(obj, self)
+      self[key] = obj
+      return obj
+    end
```
Kazuo256: This `get` function could be define outside `DB.init()` as a local function.

game/debug/view/domain_list.lua
```diff
@@ -12,23 +13,23 @@ local function sort(list)
...
-    domain[list[selected]] = nil
+    local item_name = list[selected]
+    domain[item_name] = DEFS.DELETE
     table.remove(list, selected)
     list.n = list.n - 1
   end
 
   local function newvalue(value, spec)
-    local new = spec or DB.initSpec({}, domain_name)
+    local new = spec or DB.initSpec({}, DB.loadDomain(domain_name))
```
Kazuo256: Why can't we use the already defined `domain` variable here? (see line 16)

OrenjiAkira:
done

PR #326, opened 2017-10-01
Add resource editting from devmode

OrenjiAkira:
This separates saving resources from domains, so it helps with #324, but it's not enough yet.

This also adds a new edit type in devmode, vectors. They are a vector of integers. They have all the same range and can also have a signature (which is purely aesthetic, like, calling fields 1, 2, 3 by x, y, z; it defaults to x, y, z, w)

I think we need a way to layout integer inputs smaller, and allow horizontal alignment of integer input-lists. I don't know how, but it'd be good for this. Editing animation values like this is crazy.

This also depends on #323. So merge that first.


game/debug/view/resource_list.lua
```diff
@@ -0,0 +1,71 @@
+
```
Kazuo256: Is this script too different from `debug/view/domain_list.lua`? Can we not use the same code?

Kazuo256:
> I think we need a way to layout integer inputs smaller, and allow horizontal alignment of integer input-lists. I don't know how, but it'd be good for this. Editing animation values like this is crazy.

There's `IMGUI.pushItemWidth()` and `IMGUI.SameLine()`. Maybe not worded exactly like this.

game/debug/view/resource_list.lua
OrenjiAkira: I thought the same thing but I think I forgot to refactor it. I’ll do it, it shouldn’t be too hard

OrenjiAkira:
Can we use those without heavy refacotring?

Kazuo256:
> Can we use those without heavy refacotring?

Yes, you just have to use them in `input/vector.lua` I believe.

OrenjiAkira:
I used columns with IMGUI to do the trick.

PR #332, opened 2017-10-02
Healthbar

Kazuo256:
With colors!

### DIFF
--- a/game/common/color.lua
+++ b/game/common/color.lua
@@ -1,0 +1,22 @@
+
+local COLOR = {}
+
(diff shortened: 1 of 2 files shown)


game/common/color.lua
```diff
@@ -0,0 +1,22 @@
+
+local COLOR = {}
```
OrenjiAkira: what about `domain/defs/color.lua`?

Kazuo256: It's kind of different, I guess? You think I should move it here?

OrenjiAkira: Hmm, I don't know? I guess it's ok like this.

PR #333, opened 2017-10-03
Add walltiles

OrenjiAkira:
It works, but we need to check some things.

+ Cursor image will look better if its dimensions are lesser or equal to the tile's bounds :(
+ Cursor _maybe_ should not be able to **move** to unselectable areas (like tile walls)
+ I haven't found a good celular growth thing for naught tiles
+ On that note, now the default unwalkable tile is walls and not naught tiles


Kazuo256:
For the cellular automata, we might need to use a larger window. Right now it's 3x3, but maybe 5x5 will have a better effect.

Kazuo256:
Also, you should let holes appear and expand on non-wall tiles too.

PR #341, opened 2017-10-05
Cursor only moves to valid positions

### LINKED ISSUE #340: Make cursor only walk to valid tiles
And make it try to select the most likely wanted tile in that direction

rilifon:
Closes #340 

Sometimes by logic it shouldn'1 go to certain tiles, but it feels really good so I left the way it is.

As always test before merging

### DIFF
--- a/game/view/sector.lua
+++ b/game/view/sector.lua
@@ -4,6 +4,7 @@ local RES         = require 'resources'
 local HSV         = require 'common.color'.hsv
 local SCHEMATICS  = require 'domain.definitions.schematics'
 local COLORS      = require 'domain.definitions.colors'
+local DIR         = require 'domain.definitions.dir'
 local Queue       = require "lux.common.Queue"
 
 local TILE_W = 80
@@ -19,6 +20,7 @@ local _tile_offset
 local _tile_quads
 local _batch
 local _cursor_sprite
+local _isInCone
 
 local Cursor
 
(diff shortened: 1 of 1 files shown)


game/view/sector.lua
```diff
@@ -223,11 +224,92 @@ function SectorView:setCursorPos(i,j)
...
   if not self.cursor then return end
 
-  self.cursor.i = self.cursor.i + di
-  self.cursor.j = self.cursor.j + dj
+  local sector = self.route.getCurrentSector()
+  local queue = Queue(128)
+  local pos
+  local sector_map = {}
+
+  if self.cursor.i + di < 1 or self.cursor.i + di > sector.h or  --Check if first position is inside sector
```
OrenjiAkira: I'm pretty sure `sector` has a method for checking this. If it doesn't it's better if it's there than here.

game/view/sector.lua
```diff
@@ -223,11 +224,92 @@ function SectorView:setCursorPos(i,j)
...
+    if self.cursor.validator(unpack(pos)) and
+       self.cursor.range_checker(unpack(pos)) then
+         break
+    end
+
+    --Else add all other valid positions to the queue
+    for i = -1, 1, 1 do
+      for j = -1, 1, 1 do
+        local target_pos = {pos[1] + i, pos[2] + j}
+           if target_pos[1] >= 1 and target_pos[1] <= sector.h and --Check if position is inside sector
```
OrenjiAkira: same as above

game/view/sector.lua
```diff
@@ -223,11 +224,92 @@ function SectorView:setCursorPos(i,j)
...
+  if     di == -1 and dj == 0 then   --UP
+    return (delta_j < -2*delta_i and delta_j > 2*delta_i)
+  elseif di == 0 and dj == 1 then   --RIGHT
+    return (delta_i < 2*delta_j and delta_i > -2*delta_j)
+  elseif di == 1 and dj == 0 then   --DOWN
+    return (delta_j < 2*delta_i and delta_j > -2*delta_i)
+  elseif di == 0 and dj == -1 then     --LEFT
+    return (delta_i < -2*delta_j and delta_i > 2*delta_j)
+  elseif di == -1 and dj == 1 then   --UPRIGHT
+    return (delta_i >= -2*delta_j and 2*delta_i <= -delta_j)
```
OrenjiAkira: I think the diagonals's inequations are off, but as you said, it works somehow. I would suggest fixing them just so we can try to see how it feels, though.

PR #355, opened 2017-10-07
Add def dice to bodies

OrenjiAkira:
This adds defence attribute to bodies.

Defence works based on dice. Each type of body has a different type of die (i.e. 4 sides, or 6 sides), and a `base_def`. The damage taken is decreased by `(def * dicethrow(def_die))`, in which `def` is `base_def` plus bonuses, and `def_die` is a constant body attribute, saved on its spec. `dicethrow` is a function that gives a random integer number between 1 and the argument passed.

`base_def` and `def_die` are stored as part of the body spec. bonuses are saved on the body state.

I also took the liberty of adding some things to the DB to rebalance things with this new feature.


game/domain/actor.lua
```diff
@@ -261,6 +261,7 @@ function Actor:degradeWidget(trigger, slot)
         card_info.spent = card_info.spent + 1
       end
       if card_info.spent >= card:getWidgetCharges() then
+        print(("%s has broken"):format(card:getName()))
```
Kazuo256: Leftover print

game/domain/body.lua
```diff
@@ -61,8 +65,19 @@ function Body:setHP(hp)
   self.damage = math.max(0, math.min(self:getMaxHP() - hp, self:getMaxHP()))
 end
 
+function Body:getDefence()
```
Kazuo256: Too late to start using british english.

game/domain/body.lua
```diff
@@ -61,8 +65,19 @@ function Body:setHP(hp)
...
+function Body:getDefenceDie()
+  return self:getSpec('def_die')
+end
+
 function Body:takeDamage(amount)
-  self.damage = math.min(self:getMaxHP(), self.damage + amount)
+  local defroll = RANDOM.rollDice(self:getDefence(), self:getDefenceDie())
+  local dmg = math.max(math.min(1, amount), amount - defroll)
+  self.damage = math.min(self:getMaxHP(), self.damage + dmg)
+  print(("%s took %d of damage."):format(self:getSpec('name'), dmg))
```
Kazuo256: Another leftover print

game/domain/body.lua
```diff
@@ -61,8 +65,19 @@ function Body:setHP(hp)
...
+end
+
+function Body:getDefenceDie()
+  return self:getSpec('def_die')
+end
+
 function Body:takeDamage(amount)
-  self.damage = math.min(self:getMaxHP(), self.damage + amount)
+  local defroll = RANDOM.rollDice(self:getDefence(), self:getDefenceDie())
+  local dmg = math.max(math.min(1, amount), amount - defroll)
```
Kazuo256: I think you can substitute the innermost `min(...)` for `1`*, because if `amount < 1` then `amount - defroll < 1` as well (because `defroll > 0` always).

*Actually, wouldn't zero be better? Thus allowing damage to be fully blocked, possibly causing zero damage?

OrenjiAkira: I did this exactly so that when damage is zero, it stays as zero. But otherwise the minimum damage is 1. While playing I noticed rolling zero a little too often, so I put the min to 1 to try and balance that.

But we can balance this by making attacks stronger too.

Kazuo256: I think that since we do not have a mechanic for "evading" or "dodging", the possiblity of recuding damage to zero through common means (i.e. without immunity) will be missed.

Ideally, though, we should probably discuss and test this further, since you felt it would be better this way. You can leave it like that, and I'll open an issue to investigate this later.

OrenjiAkira: i felt like i was playing tibia tbh. i was relentlessy shooting the enemy and neither of us were taking any damage in 3 out of 4 shots. i guess the attack and life stats are kinda low too, so there's that. This is all about balancing, and well, i just hate the idea of something being completely invincible just because it's high in def. Lots of RPGs don't allow zero damage unless there is immunity involved too (FF series in general comes to mind).

In the end, given other factors, dealing just one in damage is never worth it because you gon' get killed before you can beat the monster, which is clearly too much for you. But on the other hand, when there is a mob of weaklings attacking you, a ton of 1dmg might actually be worrisome.

Kazuo256: Yeah, it's about balancing in the end, and I see what you mean.

I said to leave it like that, but forgot to approve the changes. Will do that, then we can do some fun sciency design experiments someday to change it if necessary.

Kazuo256: Oh, damn, I forgot you can still remove that `min(...)`part, even reducing damage to 1.

PR #358, opened 2017-10-08
Action menu

### LINKED ISSUE #266: Action menu
There should be a button that opens a menu for the player to pick an action.

The menu could be a ring around the avatar which turns with LEFT and RIGHT. The menu items are icons hopefully.

Actions that should appear in the menu (PRIMARY ACTION uses the CONFIRM button):
1. Use Widget
2. Play Card
3. Draw Hand
4. Open Booster
5. Idle

### LINKED ISSUE #267: Ring menu
Gamestate + view

Kazuo256:
Close #266
Close #267 

Controls are reduced now:

+ F: unchanged
+ D: unchanged
+ S: primary action
+ A: action menu

And everything else is gone, save for the status screen.


Kazuo256:
Ready to go

OrenjiAkira:
I haven't reviewed yet, but test-playing made me feel weird about the controls (I tried going up and it went down, I'm probably thinking it's inverted--maybe that should be an option, for the player's preference).
But the worst thing is the lack of names for the things. We should name the items in the menu with text, please. That's good design--believe me. Icons only make sense because we attribute meaning to them, or rather, because we make them become symbols. They are not symbols until the player _knows_ what they mean. So it's okay to use text. If that's too much to do in this PR, though, you can leave it like that and we open an issue

Kazuo256:
Added labels to menu options.

For the direction movement, I followed the same principle from the buffer picker: if you press a direction, it's because you want the option that lies ahead *in that very direction*. But it's quite easy to change.

Kazuo256:
Add a few tweaks to the transparency.

OrenjiAkira:
We should open an issue about returning to the action menu after canceling a submenu, also

PR #365, opened 2017-10-08
Refactor font view helper

OrenjiAkira:
Make it so it allows getting attributes in a consistent way using the cache always.

### DIFF
--- a/game/view/actor.lua
+++ b/game/view/actor.lua
@@ -83,7 +83,7 @@ function ActorView:draw()
   if not actor then return end
   local cr,cg,cb = unpack(COLORS.NEUTRAL)
 
-  _font.set()
+  _font:set()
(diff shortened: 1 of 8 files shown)


Kazuo256:
I thought you were going to add functions to the `FONT` helper that provided font attributes directly from their ID and size. Why did you choose to make a wrapper for LÖVE's font instead? Also, in that case, why not just use a metatable? It seems the only important method is `set()`, all others could be `__index`'d back to LÖVE's font object.

OrenjiAkira:
I _was_ doing that, but it made passing arguments much more beaurocratic. I thought briefly of using a metatable, but I wasn't sure how I'd do it. I think I have to implement the __index method instead of just using the Font object. I'll tweak it.

OrenjiAkira:
done

Issue #376, opened 2017-10-09
Three actual monsters


Kazuo256:
Accepting ideas.

rilifon:
Omar, the Destroyer.


But for reals, I think is better to think of monsters if you provide some documents or related material to your universe for inspirarion
(leaks for first chapters yay or nay?)

OrenjiAkira:
I have three easy to add monsters soon to be done

OrenjiAkira:
But we def could have an omar newbie killer chara too. It should be a humanoid dog

rilifon:
or a cone

Rewasvat:
I SAW THAT!

Just a dog, no furries please. Maybe with sea-related powers? Just brainstorming.

Kazuo256:
> Just a dog, no furries please

How to dig your own grave.

OrenjiAkira:
honey u are def getting ur fursona in the game now

OrenjiAkira:
+ [ ] Slime
+ [ ] Floating Eyeball
+ [ ] Imp

Issue #97, opened 2016-08-25
Show monster's name with their sprite


OrenjiAkira:
Also add names to the monsters themselves.

Kazuo256:
What about showing it above them? After all, the life bar is above them.

Kazuo256:
Bump.

OrenjiAkira:
I was working on it, I think? I think I got side tracked. Anyway, it should be easily done. The bodyview already uses info from a meta file that comes with the sprite. It should load a name from there. All that is left is making the name visible.

Kazuo256:
What I wanted to know was about where the name should go, actually =D

rafaeljpc:
It's mandatory to show a enemy name?
Maybe a card that do that..... only boss I think is useful to know the name

Kazuo256:
I think it gives somes substance to the lore. But I see it would be a rather minor feature in the game. Maybe we can postpone this issue to prioritize other things.

OrenjiAkira:
I think it is very important for the monsters to have names. But, as I think further on the subject, I also think that they don't have to be visible along with the lifebar. I intended to make a hud that shows info on actions and monsters and such. Maybe it could, for example, show on the monster's status window.

Kazuo256:
Closed via #382

PR #382, opened 2017-10-13
Add new character and stuff

### LINKED ISSUE #324: Allow DB saving per-object

### LINKED ISSUE #337: Add tilemap resource

OrenjiAkira:
Ok, before this PR becomes too big, I'm sending it in.

This PR does the following:

1. Add a new monster sprite (#376)
2. Adds a new monster actor+body
3. Add a name display when targetting bodies (#97)
4. Adds a tileset entry in the DB ( Close #337 )
5. Makes it possible to save just one file at a time (finally! Close #324 ) 
6. It also tweaks a few things here and there in the devmode.
7. Makes the sectorview draw all floor tiles first, and then do walls + bodies on a per line basis. This fixes a minor visual glitch in which bodies clip the floor when walking upwards.

Doing this I realized how tilesets are much more detailed in how we use tiles and their quads than all animations so far. So maybe I should refactor sprites to have less info on animation frames, assuming stuff like "all quads have the same size and same offset". This would make filling animation DB entries easier and cleaner. While animations are more likely to follow that format, tilesets aren't, so it makes sense.


Kazuo256:
Haven't read through the PR yet, but let me see if I understood:

> Doing this I realized how tilesets are much more detailed in how we use tiles and their quads than all animations so far. So maybe I should refactor sprites to have less info on animation frames, assuming stuff like "all quads have the same size and same offset". This would make filling animation DB entries easier and cleaner. While animations are more likely to follow that format, tilesets aren't, so it makes sense.

You mean Sprites now all have fixed sizes and hotspots for quads, and tilesets will have the more versatile parameters? Is this included in this PR?

OrenjiAkira:
I meant we _should_ do things that way, but no, I do not implement this yet in this PR

Issue #369, opened 2017-10-09
THE GREAT RETCON


Kazuo256:
So.

Talking to @OrenjiAkira and realizing how everything's become very difficult to manage (or is it?), we formulated a HUGE AND CRUEL SIMPLIFICATION of the game as follows:

1. Actors only have ONE buffer
2. Using cards puts them on the DISCARD PILE instead of destroying them
3. You can only consume cards directly from your discard pile
4. Opening a pack directly puts all opened cards in the discard pile

Consequences:

1. Drawing a hand becomes trivial
2. Opening packs becomes trivial
3. Since cards are no longer destroyed by use, there's no need for infinite basic packs
4. We can have a menu option "Consume", which is the only way to get rid of cards

What do you think, @rilifon?

marcheing:
What is happening!?

Kazuo256:
We are going crazy is what is happining.

rilifon:
I need to ponder this while sipping from dark tea. I will bring my thoughts in the near future. May God have mercy on our souls.

Kazuo256:
Is it alright to close this by now?

The only thing we haven't adopted is the simplification of opening packs, since, as @rilifon argued, it makes things easier for the player. Other than that, we just don't have non-basic packs, but that is something we would do anyway.

PR #385, opened 2017-10-17
RETCON Part 1: Refactor actor to have only a single buffer

OrenjiAkira:
First part of #369 

1. Actors now have only one buffer. Used cards go back to backbuffer, including half-used widgets.
2. Upgrade cards, however, do not come back, and neither do cards that have the "one time only" option checked.
3. Cards are now _actual cards_ from the moment they enter your buffer, not just a specname. This allows widgets to have state of usage after being restashed.
4. Also 'remember' is now called stash, but that can be easily changed too.

I must say, from what I have tested, this improved significantly the speed and efficiency of using cards + renewing hand, but that's just my personal impression.


game/domain/card.lua
```diff
@@ -13,14 +13,12 @@ function Card:init(specname)
...
 
 function Card:loadState(state)
-  self:setId(state.id)
   self.specname = state.specname
   self.usages = state.usages
 end
 
 function Card:saveState()
   local state = {}
-  state.id = self:getId()
```
Kazuo256: Why aren't we saving the card ID anymore?

game/view/actor.lua
```diff
@@ -28,7 +28,7 @@ local function _initGraphicValues()
   _exptext = "EXP: %d"
   _statstext = "STATS\nATH: %d\nARC: %d\nMEC: %d\nSPD: %d"
   _depthtext = "DEPTH: %d"
-  _buffertext = "BUFFER #%d: %d cards [%d remembered]"
+  _buffertext = "DECK: %d cards [%d in discard pile]"
```
Kazuo256: The in-game term for "deck" is still "buffer", right?

game/domain/card.lua
OrenjiAkira: because we never have even given cards ids, plus they have zero use whatsoever. i can put it back in but the cards are going to save nothing still

game/view/actor.lua
OrenjiAkira: Oh, right. I should fix that. I might have done this without thinking

game/domain/card.lua
Kazuo256: I see. Seems fine to leave them without IDs then.

PR #393, opened 2017-10-20
Fix camera centralisation and pixel snap issues

OrenjiAkira:
The PR no one asked for

### DIFF
--- a/game/common/camera.lua
+++ b/game/common/camera.lua
@@ -1,0 +1,28 @@
+
+local Camera = require "steaming.extra_libs.hump.camera"
+local math = require 'common.math'
+
+
+local _width, _height = love.graphics.getDimensions()
+local CAM = Camera(_width/2, _height/2, 1, 0, Camera.smooth.damped(5))
+
+function CAM:attach(x,y,w,h, noclip)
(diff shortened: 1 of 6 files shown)


Kazuo256:
Was the solution to use CAM:lockPosition()? Why did you have to re-implement CAM:attach()?

OrenjiAkira:
The `lockPosition()` method does exactly the same as moving the camera to where we want with smoothness. I just thought it would be simpler to delegate the animation to hump.

It’s not the solution, however. The solution was to reimplement the attach method so it would round the translate values _only_ when actually translating, but not the position vector itself (which stays as a float value).

To be honest, though, I got the solution from [this](https://github.com/nkorth/hump/commit/a98c04e9e077ad5200207cd287be576ce599cf6b), which came up when I was reading [this thread](https://www.google.com.br/amp/s/amp.reddit.com/r/love2d/comments/6zpse1/hump_camera_making_my_game_blurry/)

OrenjiAkira:
Also since I do not have a computer that reproduces the glitch, please test it yourself so I know that it’s really fixed.

Kazuo256:
It looked fine here.

PR #407, opened 2017-10-28
Refactor pack view and maneuver actions

### LINKED ISSUE #359: Idle action wastes no time
Right now "waiting" does nothing, literally. It doesn't pass any amount of time, and it should.
But should it go tick by tick, or by time unit? If it's by time unit, we should rebalance all action's time costs (which we should anyway)

OrenjiAkira:
Now open pack menu and manage back buffer menu both share the same view.
And also, I refactored the maneuver actions. I had to fix some things here and there for it to work. 

1. Now cards need a PP reward field, for instance.
2. `_useAction` in `user_turn.lua` now checks for different cards types instead of just `art` types.
3. I also renamed some basic actions to have a clearer name, and also to have the same name as the maneuver files they correspond to.
4. Sorry @rilifon, I ended up deleting the pack view you made, but I savaged the timers.
5. This was hard. Jesus. Please tell me if anything breaks. But either way this is but part one of a great refactoring.

Also closes #359, miraculously


OrenjiAkira:
also close #359, miraculously

PR #408, opened 2017-10-29
Endless widgets + lean action system

Kazuo256:
Part of #404

Ths PR introduces two major changes:

## Lean action system

Essentially, `ACTION = MANEUVER + MAYBE(ABILITY)`

There are exactly ten possible actions in Backdoor now:

1. Idle
2. Move
3. Interact
4. Signature ability (old "primary action")
5. Draw new hand
6. Play a card (whatever it is) (possibly activates an art ability)
7. Activate a widget ability
8. Stash card
9. Consume cards from backbuffer
10. Open a pack

They are now all described in the maneuvers in `domain.maneuvers`. Every maneuver has:

1. A `param_specs` table describing the basic parameters they need
2. A `activatedAbility` getter for actions that may activate one (art cards, signatures, etc.)
3. A `validate` method to check parameter integrity
4. A `perform` method that actually does the deed

A performed maneuver may or may not activate an ability, hence the formula in the beginning of the section.

All in all, the action system code is now GODDAMN SIMPLER. Just look at that line diff count of the PR.

## Unlimited Widget Works

The slots for widgets in an actor are now limitless. Placement restrictions are still in place though.

This means that playing a widget card no longer requires a prompt to choose a slot — it just gets appended at the end of the actor's widget list. IF it has a placement conflict, the previously present widget of matching placement is discarded.

I haven't fixed the widget picker menu yet, since I didn't want to bloat this PR any further. It doesn't pose a threat for now anyway, but it should be done as soon as this PR gets merged.


game/domain/maneuver/use_signature.lua
```diff
@@ -0,0 +1,24 @@
...
+end
+
+function SIGNATURE.validate(actor, sector, params)
+  return ABILITY.checkParams(actor:getSignature().ability, actor, sector,
+                             params)
+end
+
+function SIGNATURE.perform(actor, sector, params)
+  local signature = actor:getSignature()
+  actor:spendTime(signature.cost)
```
OrenjiAkira: shouldn't this award PP as well?

game/domain/maneuver/draw_new_hand.lua
```diff
@@ -1,20 +1,23 @@
...
+end
+
+function DRAWHAND.validate(actor, sector, params)
   return actor:isHandEmpty() and actor:getPP() >= DEFS.NEW_HAND_COST
 end
 
-function maneuver.perform(actor, sector, params)
-  actor:spendPP(DEFS.NEW_HAND_COST)
+function DRAWHAND.perform(actor, sector, params)
   for i = 1, DEFS.HAND_LIMIT do
```
OrenjiAkira: here, why did you remove, the `spendPP` line?

Kazuo256: Just an accident I'd guess.

Kazuo256:
There, fixed. Good thing you spotted all that >.<

Kazuo256:
Also, what do you think could still be improved?

OrenjiAkira:
I think the `user_turn` gamestate could be leaner, basically. But as it is, it shouldn't need much changing from now on.

Kazuo256:
Yeah, refactoring `_userTurn()` I realized you could be much simpler. Will try to do that on the go with the following issues.

Issue #353, opened 2017-10-06
Remembering card should buff corresponding attribute


Kazuo256:
Until next card is played or hand is finished.

PR #417, opened 2017-10-31
Widget buffs

### LINKED ISSUE #244: Add Attributes Temporary Modifier and Effects
Think a clever way to implement temporary modifiers to attributes, and create effects to apply them

 My idea:

Temporary attributes affect directly the attrib modifier variable, but the actor stores "timers" or "clocks" that indicates changes in future turns, using countdowns. For example, giving a +3 to ATH for 10 turns would increase the ath_mod variable by 3, but the actor receives a static effect "reduce ATH by 3" with a countdown for 10 turns.

The game will update all countdown effects in the beguinning of the actor turn, and check if anyone has reached 0. If so, apply the effect.

Opinions?

### LINKED ISSUE #404: Maybe use widgets for EVERYTHING
An indecent proposal:

What if there were no temporary state in actors/bodies other than WIDGETS?!

Buffs, poison, equips, locks, curses, everything could be represented through widgets!

Of course, this would mean removing the limit of widgets a character can "carry", save for equipments (which are already limited by the "placement" mechanic anyway).

Kazuo256:
Allow widget cards that operate on attributes (i.e. buffs). Works only with "additive" bonuses and turn-triggered widgets (they lose charge per turn).

Can we consider this to close #244, although it still doesn't support "set" bonuses?
Close #404, because the foundation is laid.

This PR adds three new cards, the "Stash Bonus" cards. They are supposed to be spawned when a card is stashed when #353 is implemented. Also, I reduced card copies in the initial buffer to 2, since it was getting too big.


OrenjiAkira:
It's working but I have some questions:
1. Why don't `gun` and `sord` have a `trigger` value?
2. What is the diff between `on_tick` and `on_turn`?

Kazuo256:
> Why don't `gun` and `sord` have a trigger value?

At first, because they already lose charges when used. But I agree that the most consistent would be to have that work through the trigger mechanics I implemented too. It's working the way it is, so I thought of adjusting this in another PR.

> What is the diff between `on_tick` and `on_turn`?

`on_turn` trigger at the beginning of an actor's turn (which means it never triggers for bodies without actors). `on_tick` *would* trigger every N turn ticks or so, regardless of actor turns (and would also affect bodies), but I did not implement the trigger in this PR (also wanted to do later). Essentially, if you are a fast character, `on_turn` triggers more often to you, while `on_tick` triggers the same for every one.

I decided to go with `on_turn` for the stash bonuses because it sort of guarantees that you'll have a fixed number of chances of making use of the bonus. If you are slowed by whatever reason, an `on_tick` bonus would potentially go to waste.

PR #421, opened 2017-10-31
View/improve widgetview

OrenjiAkira:
This improves the widgetview so that it displays how many charges are left and what kind of trigger it uses for spending. It adds also a validation of the widget view state, so that it doesn't open if there are no widgets 'equipped'.

It also adds an armor widget for testing.


game/domain/params/choose_widget_slot.lua
```diff
@@ -11,7 +11,7 @@ function PARAM.isValid(sector, actor, parameter, value)
   if not actor:getBody():hasWidgetAt(value) then
     return false
   end
-  return true
+  return not not actor:getBody():getWidget(value):getWidgetActivation()
```
Kazuo256: This only works when we choose an widget to activate. It won't work for ablities that require a target widget to destroy, for example.

game/view/pickwidget.lua
```diff
@@ -49,23 +52,31 @@ function PickWidgetView:draw()
...
+  end
+  for index, info_str in ipairs(strs) do
     local selected = self.selection == index
     if selected then
       g.setColor(0xff, 0xff, 0xff, self.alpha*0xff)
     else
       g.setColor(0x16, 0x16, 0x16, self.alpha*0xff)
     end
-    g.rectangle("fill", 0, 0, _text_width, _BLOCK_HEIGHT)
+    g.rectangle("fill", 0, 0, width+8*_PADDING, _BLOCK_HEIGHT)
```
Kazuo256: This makes the rectangle fir to the text. In particular, if you only have a "Sord", it becomes quite small. Is this intended? (just checking, I'm OK either way)

game/domain/params/choose_widget_slot.lua
OrenjiAkira: yes but for now it makes the menu not close when i select an invalid widget

game/view/pickwidget.lua
OrenjiAkira: yes, this also makes the rectangle fit for larger names, like the stash buffs

game/domain/params/choose_widget_slot.lua
Kazuo256: Ok, I'll make an issue for when we have abilities that target widgets

OrenjiAkira:
about the armor, i thought you were going to still implement the armor usage, so idk how we would go about "spending" it

Kazuo256:
It would have to be a different trigger, something like "On defend" or "On hit taken".

OrenjiAkira:
oh, that means that widgets can only have _one_ kind of trigger. I think that's fine, but we should be careful balancing stuff like this, and not accidentally make an armor that casts long distance really strong magic and only degrades with hits. you'd be invincible, almost

OrenjiAkira:
do you want me to implement "on_defend" this trigger? This PR was just for the view, we can do it in another PR

Kazuo256:
> do you want me to implement "on_defend" this trigger? This PR was just for the view, we can do it in another PR

Yeah, just change to `on_tick` or `on_turn` for now.

> that means that widgets can only have one kind of trigger

Haven't thought of giving them multiple triggers. Sounds interesting. Will open an issue.

OrenjiAkira:
done

Issue #308, opened 2017-09-27
Collectable packs


Kazuo256:
In addition to basic packs per background, there must be collectable packs dropped by creatures or found lying around. Once collected, actors should be able to open them just like basic packs.

PR #429, opened 2017-11-07
Make packs behave like loot

OrenjiAkira:
With this PR, packs now behave like loot. Opening a pack has a similar behaviour of getting a monster loot: you have chances of getting specific items of specific categories. Basically [this](http://journal.stuffwithstuff.com/2014/07/05/dropping-loot/).

First part of #308. Monsters still do not drop packs right now. @Kazuo256 is working on refactoring `gamestate/user_turn`, so I have decided to wait for now. But with this PR it should be a piece of cake. All one needs to do is call `gamestate/open_pack` passing the dead monster's actor object as the argument, or even just their collection's name.

The downside of this PR is that opening a pack from the action menu has a chance of not yielding cards at all. This translates to the action menu just closing instead of opening a pack.


game/domain/pack.lua
```diff
@@ -1,16 +1,20 @@
...
+  local collection = DB.loadSpec('collection', collection_name)
   local pack = {}
-  local total = #cards
-  for i=1,DEFS.PACK_SIZE do
-    table.insert(pack, cards[RNG.generate(total)].card)
+  local n = 0
+  for _,card_drop in pairs(collection.cards) do
+    local p = RANDOM.generate(1, 100)
+    if p <= card_drop.drop then
+      n = n + 1
```
Kazuo256: You can adapt this for to repeat until `PACK_SIZE` cards have been selected.

The key is considering that a drop chance doesn't represent whether a card from that set will or will not appear, but rather that it's the chance a card from the set will appear *instead of one of the others*. That is, you can "sum" all drop chances of a collection, in a way that you can always request exactly *one* card from it, and exactly one of the possible sets will be picked from. Then, you repeat that `PACK_SIZE` times.

OrenjiAkira: I can do that but I'm not sure this is the most fun way of doing things. I'll leave packs with 5 cards, but I think it's more _fun_ if the player receives a random ammount of cards. Also, receiving always 5 cards means that you will constantly get around the same amount of exp, which makes the game a little unbalanced and generates less feelings of progression

Kazuo256: Oh, so that's why you did it that way. I thought the problem was the implementation, not the design.

I don't see much problem with varying pack sizes. In fact, I hadn't even considered it, since all card games come with fixed pack sizes. But then again, our cards are loot, and every dungeon crawler has "varying size loot", so it also makes sense.

I'll leave it up to you what you prefer then. If you want to revert the last commit (the one that affixates the size), just force-push the previous commit onto this branch.

OrenjiAkira: Ok

PR #440, opened 2017-11-09
Add initial buffers to specific backgrounds

OrenjiAkira:
Now arcanists only have spells on their decks, craftsmith starts with guns, and brawlers have sords


OrenjiAkira:
Added turn-based self-activated abilities in widgets

Kazuo256:
Please make all initial buffers have the same size

PR #443, opened 2017-11-10
Aptitudes

### LINKED ISSUE #366: Feature proposal: attribute aptitude
Just a crazy idea to think in the future.

Actor specs can have an "attribute aptitude" instead of "attribute base", and actor states have "attribute experience", making the actual attribute value something like "experience/aptitude". This stops players from making muscular mages or barbarian geniuses (which is debatably undesirable).

It also allows all actors to start with the same "base attributes", and it naturally adjusts to their aptitudes. For instance, all actors could start with 100 "attribute experience" on all attributes, and actors with aptitude 10 would have a grater starting value than actors with aptitude 20 (we could invert the values when showing the player for clarity).

Anyway, not sure if this will be any good. Let's wait until we have a proper environment for playtesting.

Kazuo256:
Close #366


OrenjiAkira:
Awaiting conflict resolution before approval

Issue #467, opened 2017-11-23
Arcanist Basic Pool Cards ver 1.0


rilifon:
Make the following cards for the Arcanist background card pool

**================================**

- [x] Firebolt
- **Art** (Arcana)
- Choose a target body. It takes damage

**================================**

- [x] Fireball
- **Art** (Arcana)
- Choose a valid tile. Create an AoE (#457) that damages every body inside it

**================================**

- [x] Teleport
- **Art** (Arcana)
- Choose a target valid position. Teleports actor (and its body) to it.

**================================**

- [x] Blind
- **Art** (Anima)
- Puts a blind widget on target actor, that for X turns will debuff his FoV range

**================================**

- [x] Heal
- **Art** (Anima)
- Choose target body, heals hp for some amount

**================================**

- [x] Focus
- **Widget/Tool** (Arcana)
- Buffs ARC for some amount 
- Has X charges. Spend one every time this actor uses an Arcana Art card.

**================================**

- [ ] Grimoire
- **Widget/Tool** (Arcana)
- When cast, choose a target Art not one-use-only card in your hand ( #458). Destroys that card and store its information on the Grimoire.
- Has X charges, and one use, spends one and creates a copy of store card and put it in actors hand.

**================================**

- [x] Haste
- **Widget** (Anima)
- Buffs actor's speed
- Has X charges, spends one every turn

**================================**

- [ ] Leech
- **Art** (Arcana)
- Choose a target body
- Damages it and heal this actor on amount of damage given

**================================**

- [ ] Sense
- **Widget** (Anima)
- Shows not visible monsters (for now, outside actors FoV)
- Has X charges, spends one every turn

**================================**

- [x] Divination
- **Art** (Arcana)
- Draw 2 cards.

**================================**

- [x] Sticky Flame
- **Art** (Arcana)
- Creates a widget on target body named "Sticky Flame". Widget has X charges, and spends one every turn. When widget is done (#459), explodes and damages target body.

**================================**

- [ ] Echo
- **Art** (Arcana)
- Choose another target Art card in actor's hand. Cast a copy of it.

**================================**

- [x] Soul Robe
- **Widget/Suit** (Anima)
- Buffs actor ANIMA

**================================**

Kazuo256:
Closing but moved missing cards to separate issues.

PR #508, opened 2017-12-07
Card/blind heal

rilifon:
Added cards Heal and Illusion of Darness (Blind) for Arcanist.

Also made some changes on cards descriptions for more consistency

References #467


Kazuo256:
Corrupted branch

Issue #465, opened 2017-11-23
Brawler Basic Pool Cards ver 1.0


rilifon:
Make the following cards for the Brawler background card pool

**================================**

- [x] Laser Spear
- **Widget/Weapon** (Corporis)
- On use, damages target body (melee range)
- Has N charges, spends one on every use

**================================**

- [x] Spark Rod
- **Widget/Weapon** (Corporis)
- On use, damages target body (melee range) then creates an AoE (#457)  that damages all bodies within that range (center of range is on first body)
- Has N charges, spends one on every use

**================================**

- [x] Rush
- **Art** (Corporis)
- Choose a direction (#90), then runs in that direction until it hits a body (damages that body) or a wall

**================================**

- [x] Rage
- **Widget** (Anima)
- N charges, spend one every turn
- Buffs ATH and SPEED
- Debuffs DEF

**================================**

- [x] Calming Down
- **Widget** (Anima)
- Regen Health each turn
- 1 charge. Is spent if actor uses any action besides Idle

**================================**

- [x] Power Suit
- **Widget/Suit** (Corporis)
- Buffs DEF
- Debuffs SPEED

**================================**

- [x] Super Jump
- **Art** (Corporis)
- Chooses a valid tile with no bodies present. Teleports to that tile, creating a damage AoE (#457)  on nearby bodies

**================================**

- [x] Stomp
- **Art** (Anima)
- Creates an AoE (#457) that damages nearby bodies and gives them the Stun widget (#462, #464)

**================================**

- [x] Dropkick
- **Art** (Corporis)
- Choose a nearby body (melee-ranged) and damages it
- Actor who used this art receives some recoil damage

**================================**

- [ ] Shove
- **Art** (Corporis)
- Choose a nearby body (melee-ranged), it takes some small damage and is shoved (#463) a few tiles back

**================================**

- [ ] Reckless Strike
- **Art** (Anima)
- Damages target body (melee-ranged) with great damage, but has 50% chance of failing

**================================**

- [x]  Seedling Sap
- **Widget/Tool** (Anima)
- Has X charges. On use, spends one charge, actor draws 2 cards and receives the widget "Tipsy"
- *Tipsy*: Has Y charges. Spends one every turn. Debuffs actor SPEED.

**================================**

Kazuo256:
Closing but moving missing cards to separate issues.

PR #513, opened 2017-12-08
Card/trivial cards

rilifon:
Added card Calming Down (old Meditate) dor Brawler (#465)

Added cards Soul Robe and Stick Flame for Arcanist (#467)

For trivial cards still need to make Reassemble, but will do at another PR.

NOTE: For card Calming Down, had to add 2 charges instead of 1, since it will spend the first one when you cast the card (using on_act trigger). We should fix this in another commit so we can make it with only one charge

### DIFF
--- a/game/database/domains/actor/arcanist.json
+++ b/game/database/domains/actor/arcanist.json
@@ -5,25 +5,13 @@
   "cor":-4,
   "initial_buffer":[{
       "amount":1,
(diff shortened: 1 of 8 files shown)


OrenjiAkira:
I'm gonna accept this, but please make card descriptions shorter. They must not go beyond 6 lines on screen when viewed on the hand.

PR #516, opened 2017-12-10
Set onetime-only flag to some widgets

OrenjiAkira:
Like blind, sticky flame, tipsy.
Stuff that shouldn't go back to your buffer, lmao.

Also added a breakline filter to cardinfo displays

### DIFF
--- a/game/view/helpers/card.lua
+++ b/game/view/helpers/card.lua
@@ -101,7 +101,7 @@ function CARD.drawInfo(card, x, y, width, alpha)
   g.translate(0, _title_font:getHeight())
 
   _text_font.set()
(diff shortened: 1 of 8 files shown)


game/view/helpers/card.lua
```diff
@@ -101,7 +101,8 @@ function CARD.drawInfo(card, x, y, width, alpha)
   g.translate(0, _title_font:getHeight())
 
   _text_font.set()
-  g.printf(card:getDescription() or "[No description]", 0, 0, width)
+  local desc = card:getDescription():gsub("[^\n][\n][^\n]", " ")
```
Kazuo256: This doesn't work in Fireball's description: it eats the characters before and after the breakline.

OrenjiAkira:
done

Issue #472, opened 2017-11-23
Gadgeteer Basic Pool Cards ver 1.0


rilifon:
Make the following cards for the Gadgeteer background card pool

**================================**

- [ ] Jetpack
- **Widget/Suit** (Arcana)
- Buffs SPEED
- Adds tag "Float" (#460) on actor, so he can walk over "holes" and over any grounded AoE
- Has X charges. Spends one every turn

**================================**

- [ ] Radar
- **Widget/Tool** (Arcana)
- Marks the exit on the minimap (while widget is active)
- Has X charges, spends one every turn

**================================**

- [x] Goggles
- **Widget/Tool** (Corporis)
- Buffs player FoV range
- Has X charges, spends one every turn

**================================**

- [x] Wrench
- **Widget/Weapon** (Corporis)
- On use, damages nearby body (melee range) based on how many drones you have nearby
- Has X charges, spends one every use

**================================**

- [x] Blasters
- **Widget/Weapon** (Arcana)
- On use damages target body (long range)
- Has X charges, spends one every use

**================================**

- [x] Reassemble
- **Art** (Corporis)
- Choose target drone, and moves it to target valid location

**================================**

- [x] Synch
- **Art** (Arcana)
- Gives a widget to target Drone that buffs DEF

**================================**

- [ ] Drone: Ion Cannon
- **Art** (Arcana)
- Creates the body Ion Cannon on target valid position (#461)
- *Ion Cannon*:
  - Has the "on interact" (#470) widget: Has X charges, on activation spends one charge and creates a ion laser beam that shoots in opposite direction the actor has used. After use if widget has no charges, destroys the Ion Cannon body

**================================**

- [x] Drone: Turret
- **Art** (Corporis)
- Creates the body Turret on target valid position (#461)
- *Turret*:
  - Has the widget: Has X charges. On every turn, if there is a nearby non-player body (with an actor), damages it, spending one charge. If it has no charges left, destroys the Turret body

**================================**

- [ ] Drone: HealBot
- **Art** (Arcana)
- Creates the body HealBot on target valid position (#461)
- *HealBot*:
  - Has the "on interact" (#470) widget: Has X charges, on activation spends one charge and actor who activated it. After use if widget has no charges, destroys the HealBot body

**================================**

- [ ] Drone: Force Field Beacon
- **Art** (Arcana)
- Creates the body Force Field Beacon on target valid position (#461)
- *Force Field Beacon*:
[...]

PR #517, opened 2017-12-10
Operators that detect nearby bodies

Kazuo256:
Add operators `count_nearby_bodies` and `find_nearest_body`, besides the Wrench and Sensor Bomb card from #472.

Supposedly, tihs would allows us to also implement the Turret card, but there is one problem. A deployed Turret requires two widgets, or a widget with two triggers, since it triggers every tick (to fire) *and* when is loses all charges (to destroy itself).


game/database/domains/action/MELEE.json
```diff
@@ -11,17 +12,18 @@
       }],
     "params":[{
         "typename":"choose_target",
-        "max-range":1,
+        "body-only":{
+          "body-type":"<none>"
```
OrenjiAkira: I think this will crash when executed. But it's my bad, I'm the one who made input enums default to that `'<none>'` string.

game/domain/body.lua
```diff
@@ -80,6 +81,25 @@ function Body:saveState()
   return state
 end
 
+--[[ Spec-related methods ]]--
+
+function Body:isSpec(specname)
+  if not specname or specname == '<none>' then
```
OrenjiAkira: Oh ok, this is weird, but ok. This solves that thing.

game/database/domains/card/sensor-bomb.json
```diff
@@ -0,0 +1,27 @@
...
+          "pos":"par:pos",
+          "vit":50,
+          "typename":"make_body",
+          "widgetspec":"proximity-detector",
+          "def":50,
+          "bodyspec":"machine"
+        }]
+    },
+    "cost":60
+  }
```
OrenjiAkira: no playpoints?

PR #561, opened 2017-12-19
Fix camzoom bug, enhance a few inputs

### LINKED ISSUE #557: Sometimes, camera does not go back to original 1x zoom
I don't know how to reproduce it, but I suspect it's related to lag.

OrenjiAkira:
Closes #557, **probably**

Also add 'return' as available key to CONFIRM action, allow drawing new cards with ACTION_1 action, and allow closing handview with ACTION_1 as well.

### DIFF
--- a/game/database/settings/controls.json
+++ b/game/database/settings/controls.json
@@ -4,7 +4,7 @@
     "ACTION_2":["e"],
     "ACTION_3":["w"],
     "ACTION_4":["q"],
-    "CONFIRM":["f"],
+    "CONFIRM":["f", true, "return"],
     "CANCEL":["d"],
     "SPECIAL":["s"],
     "EXTRA":["a"],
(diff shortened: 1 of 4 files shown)


Kazuo256:
Did you manage to reproduce the zoom bug consistently?

OrenjiAkira:
No, but I have narrowed it down ti the fact that the zoom was only being update on the draw method if the actionmenu view. When we close that menu, we create a tween in which when it ends it makes the view’s draw method stop being called. When it lags, the zoom updates only to the last time the draw method was called, which might be well before it goes back to 1. Then the tween ends because the last update call was a long time ago, and this stops the draw method from being called. Thus the zoom isn’t updated, and it stays zoomed in. The solution was to reset the zoom value to 1 when the tween ends, instead of just updating it on the draw method

PR #568, opened 2017-12-21
Refactor playpoints

OrenjiAkira:
Now playpoints are dropped when you kill monsters.
For now monsters do not have EXP, and thus they all have the same "strength level", which makes them all drop the same amount of PP, until you become stronger than them.


game/domain/actor.lua
```diff
@@ -413,4 +413,12 @@ function Actor:getPP()
   return self.playpoints
 end
 
+function Actor:getStrength()
```
Kazuo256: I suggest "getPowerLevel" or something

Issue #373, opened 2017-10-09
Encounter difficulty


Kazuo256:
Allow creation of stronger instances of monsters.

PR #579, opened 2017-12-22
Rebalance power levels

OrenjiAkira:
Now monsters have a starter "powerlevel" which is a value you add at the encounter transformer to measure how powerful a monster is gonna be. It has a variation of 20%.

The powerlevel now uses both body and actor attributes AND the relation is now powered by 5, to make powerlevel differences much more proeminent. Now a slightly stronger monster might give you more PP. Also monsters distribute their upgrade experience based on their aptitudes.

It would now be nice to show, like the damage animations, how much exp and pp you are getting, but I haven't gotten around to do that in this PR.

Also fix bug in which activated widgets stilll gave PP.


OrenjiAkira:
Never said exponential

Kazuo256:
Oh, right. I guess that was just my expectation.

Kazuo256:
By the way: this closed issue #373

PR #586, opened 2017-12-25
Remove stash mechanic

Kazuo256:
I recommend playtesting this branch before accepting the PR.

From what I playtested, the abscence of the stash mechanic made gameplay more dynamic, although there are some moments where stashing is missed. For instance. when you have a Divination in a full hand, but can't stash some other useless card, you are forced to wait before using Divination with its full value. I don't think it's a problem, but may be worth considering.


OrenjiAkira:
Well, after some testing I have two main concerns:

1. The 'play' text on the handview changed case, and it looks weird now.
2. When you have an unwanted widget, particularly unwanted equipment, on your hand, you have nothing you can do about it and you also can't buy a new hand and it sucks

OrenjiAkira:
ok, correction: the 'play' text on the handview is my fault i need to tweak it a little later on

Kazuo256:
> 2. When you have an unwanted widget, particularly unwanted equipment, on your hand, you have nothing you can do about it and you also can't buy a new hand and it sucks

Why can't you buy a new hand?

OrenjiAkira:
Ah, I see, now you can draw new hands when you still have cards in your current hand.
Will have to test play further then

OrenjiAkira:
Ok, this seems fine, though it seems that since now you can both play a card or draw a new hand at any given time, we are short of one input (or two, if we want a quick button for backbuffer-managing.

The only thing that truly bothers me is that by making these two maneuvers independent, they are now competing for the same input in `user_turn.lua`. Choose only one to stay (preferably `PLAY_CARD`) and remove the other.

There are other things to balance though

Kazuo256:
Done

PR #597, opened 2018-03-16
Add tileset alternative as test

OrenjiAkira:
We need to add some variety to the map to make actual use of it, but I noticed walls are particularly _boring_ because of their limited height and overall rectangular cuts (It always ends in a right angle, and it looks weird). Adding more height might make it look better, but it will start blocking monsters behind walls, and then we will have to start rendering them with alpha and aaaaaaaa

Anyway, play up to sector2-type sectors and tell me what you think of these sprites.


rilifon:
Actually drawing monsters behind walls with alpha is something I'm very
interested in implementing ^^

Em Qui, 15 de mar de 2018 22:16, W. Akira Mizutani <notifications@github.com>
escreveu:

Kazuo256:
Didn't look that boring to me, think it felt very dungeon-y. But I agree that maybe the walls should be taller. It sort of seems like the place has been flooded by sand and the top of the walls are barely poking out. 

Is this PR still a WIP? Can we give review/merge it yet?

OrenjiAkira:
I guess y'all can merge this and I can patch up the walls later on.

Notice that I made alternative tiles for each type (floors/walls) so we can mix them randomly. Not sure how it would look, tho.

PR #605, opened 2018-03-22
Make ON_PLAY trigger happen BEFORE executing card effect

### LINKED ISSUE #603: Widgets that are triggered with ON_PLAY are triggered on the same turn they are played
they play themselves

image

probably a residue from #598

OrenjiAkira:
Closes #603


Kazuo256:
#610 solves this without change the trigger order

PR #618, opened 2018-03-30
Drops

### LINKED ISSUE #613: Implement "drops" maybe
"Drops" are things in tiles that when stepped on they activate an ability. They are not necessarily "dropped", as they can just "be" there. Like traps or packs or playpoints (read further).
Unlike bodies, they are _walkable_, and also _untargettable_.

The feature idea came from talking to @Kazuo256 this last reunion.
But everyone should give their two cents on it too, probably.

The whole thing started when we were talking about balancing PlayPoints, and he suggested that gathering PP should be _perpendicular_ to killing monsters. Meaning they are completely independent things. And "drops" would suffice, because we could just make collecting PP a thing you do by stepping on "drops", that can be generated in any number of ways (pre-existing, or dropped destroying bodies in general).

This would also mean refactoring packs to be drops, though that's a whole different issue.

Kazuo256:
Close #613 

TODO list:
+ [x] Auto-discard hands after N inactive cycles
+ [ ] Refactor damage/heal/pp spritefx
+ [ ] Implement simple element type for drops
+ [ ] Move generation of food to a proper transformer
+ [ ] Improve drawing of multiple drops
+ [ ] Make drops oscillate


Kazuo256:
This is a good moment to test the new feature. Both food drops and hand timeout are working (the code still needs some improvements). You can change how much PP the food drop gives by tuning its ability through devmode.

If feedback for this is good, I'll implement the rest.

OrenjiAkira:
Ok, my first complaint is monsters steal your drops. They don't even do it on purpose, they just accidentally steal it from you while chasing you.

I like that they *can* steal it from me, but I don't like that *all* of them do it. It would be more interesting if there were certain kinds of monsters who are *hungry* like slimes, thieves, etc.

Secondly, I like very much the feeling of exploring for the sake of looking for loot. It's quite fun. I think there needs to be a better balance, though – drawing cards is now even harder than before, and keeping up with the monster's power progression is hard. But maybe we'll nerf that monster progression thing later on, so that is not a big deal. Not being able to draw cards so soon as I did in my playthroughs, however, is.

OrenjiAkira:
Ok, I've tested it, and this PR goes somewhat well with #617

Kazuo256:
> Ok, my first complaint is monsters steal your drops. They don't even do it on purpose, they just accidentally steal it from you while chasing you.
>
> I like that they can steal it from me, but I don't like that all of them do it. It would be more interesting if there were certain kinds of monsters who are hungry like slimes, thieves, etc.

Yeah, there's that. We could either tweak the AI to avoid drops (though that might look weird) or make the drop-grabbing mechanic more elaborate. Something like drops can only be grabbed by certain types of... bodies? If possible, I'd like to avoid requiring user input to "pick up" drops.

> I think there needs to be a better balance, though – drawing cards is now even harder than before, and keeping up with the monster's power progression is hard

This should be easy to balance by tweaking drop spawn and how much PP food gives. I tried with food giving exactly 20 PP before setting it to 5, and the game became much easier (and annoying, because I kept drawing hands just so I wouldn't "waste" food). Try changing the PP value and see if it feels better.

Kazuo256:
Ok, I think I've got a good solution.

Since drops have abilities, we can use that ability's inputs to "check" whatever steps on the drop. Then we reject anything that does not satisfies the input conditions (the same way we do with conditional triggers).

With this, I've made the food drop be picked only when an actor that is not at full PP steps on it. I've also made everyone start at full PP, so monsters don't pick up food, and neither does the player if they have full PP.

Lastly, I increased the food PP to 10, which seems good so far. Keep testing!

OrenjiAkira:
Okay, but does this mean that monsters, when me make them use cards, will also be prone to gathering food in order to cast stuff? If we make "smart" monsters won't that mean they will actively pursue for food after an encounter in which they depleted their PP? That's.... curious. Like, eventually, monsters "recover" from their encounter with you, and that might mean they consumed the sector's natural drops....... veryyy curious.

Kazuo256:
Yes, but that is kind of inevitable if we think the other way around. If PP is required to draw a hand, and we want monsters to draw hands eventually, they will need some way to recover PP or just have a limited number of draws.

The greater problem in this case is that monsters and players share the same pool of PP. In particular, with the design from `dev` the only way monsters can recover PP is by killing stuff. Guess we should start being more careful with the monster side of mechanics.

Anyhow, *if* we stick with the drop system, there are two things I can think of that would mitigate the concurrency for food:

1. [easy] Make food only be pickable by certain species (using the conditional trigger system)
2. [not so easy] Make monsters drop food

OrenjiAkira:
Making monsters (albeit just some) drop food was kinda a thing we needed anyway, right? And were drops always supposed to be food too? (Like, I remember we discussing the possibility of drops envolving packs as well)

We could make that each monsters has a particular set of types of food they like, and they only eat from that set. That's overly complicated, but it's something I wouldn't put past a roguelike to do. Also we could make _some_ monsters naturally regenerate their PP via a permanent regen widget or something – with the trigger that it only heals vert little and only when it has completely idle turns, or something.

Kazuo256:
Yes, the idea is to make some monsters drop food, and also to make other kinds of drops (since it's a domain of its own, that's just a matter of having sprites and writing them in devmode). I was just not sure whether you were happy or not that monsters were eating our food, so I tried to explain why it was inevitable but also not really a problem (because conditional trigger can simulate the "food preference" you described and yes, we could make some of them recover PP over time too).

For now, all I need to justify the rest of the PR is that either @rilifon or @Detril playtest it too and give some feedback.

Kazuo256:
Also, gadgeteer is much more playable now. And fun XD.

Issue #551, opened 2017-12-14
Refactor sector view for readability


OrenjiAkira:
Maybe divide it into smaller views, or use helpers.

PR #623, opened 2018-04-05
Refactor colors and sector view in parts

OrenjiAkira:
This PR does two things:

+ All common colors defined in `domain/definitions/colors.lua` are now objects of type color that allow color multiplication.
+ Cleans very little of the code of sector view in order to draw tiles in another file.
+ Ok, three things: adds a method of getting visible tiles in the camera module.


OrenjiAkira:
related to #551

PR #637, opened 2018-04-05
Do not close menu if action is invalid

### LINKED ISSUE #634: When not possible to enter a menu, don't leave menu
Instead play denied sfx

Kazuo256:
Close #634

### DIFF
--- a/game/view/pickwidget.lua
+++ b/game/view/pickwidget.lua
@@ -2,7 +2,6 @@
 local RES = require 'resources'
 local DEFS = require 'domain.definitions'
 local Color = require 'common.color'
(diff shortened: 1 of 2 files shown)


game/view/pickwidget.lua
```diff
@@ -1,7 +1,7 @@
 
 local RES = require 'resources'
 local DEFS = require 'domain.definitions'
-local COLORS = require 'domain.definitions.colors'
```
OrenjiAkira: Remove this from the PR because I fixed it in another PR just for minor fixes like this.

Kazuo256: Done

PR #654, opened 2018-04-12
Improve drawing of multiple drops on same tile

Kazuo256:
Signed off by: @rilifon

Close 362


OrenjiAkira:
I'm going to edit your comment so that it properly closes the issue.

OrenjiAkira:
Actually, no, the issue you mention has nothing to do with this.

PR #681, opened 2018-04-27
Make route be generated as a graph

### LINKED ISSUE #614: Make sectors be generated on a generic graph-like pattern instead of a tree-like pattern
This would allow sectors to have "loops". Also this would mean generating the whole world when the route is first generated. This would make things...different! It would also drop the "depth" attribute.

OrenjiAkira:
Route is now generated as a static, but modifiable graph of sectors.
This is the first step to making more complex route maps. As of now, the map is static, but it can most certainly be modified procedurally using grammars and such.

This also changes how exits are saved/loaded/stored. This is because all sector's states are generated beforehand, as opposed to being generated only when needed. Their tiles, however, still maintain "lazy" generation, and are constructed only when you enter a previously unvisited sector.

Closes #614


game/domain/builders/sectors.lua
```diff
@@ -0,0 +1,132 @@
...
+  route_map:connect(z4_3, z1_2)
+  -- connect zones to seed regions
+  route_map:connect(z1_4, s1)
+  route_map:connect(z1_5, s2)
+  route_map:connect(z2_4, s3)
+  route_map:connect(z2_5, s4)
+  route_map:connect(z3_4, s5)
+  route_map:connect(z3_5, s6)
+  route_map:connect(z4_4, s7)
+  route_map:connect(z4_5, s8)
```
Kazuo256: This *could* be more data-driven

game/domain/builders/sectors.lua
```diff
@@ -0,0 +1,132 @@
...
+  first_sector.tiles = tiledata.tiles
+  first_sector.h = #tiledata.tiles
+  first_sector.w = #tiledata.tiles[1]
+  first_sector.depth = 0
+  first_sector.bodies = { pbody }
+  first_sector.actors = { pactor }
+  first_sector.generated = true
+  first_sector.exits[r_0].pos = tiledata.exit
+
+  return sectors
```
Kazuo256: Could also return first sector to facilitate user code

game/domain/transformers/bootstrap.lua
```diff
@@ -1,5 +1,8 @@
 
 local SectorGrid = require 'domain.transformers.helpers.sectorgrid'
+local BUFFER_BUILDER = require 'domain.builders.buffers'
+local Actor = require 'domain.actor'
+local Body = require 'domain.body'
```
Kazuo256: Unused requires

PR #704, opened 2018-05-11
Wall meshes

Kazuo256:
Procedurally create a mesh for walls, which handles cornes and allows future features for procedural details on the scenery.

You can remove the wall margin by changing the `_MARGIN_W` and `_MARGIN_H` variables in `view/sector/wall.lua`. I left the margins because I think with smooth FOV it won't look so bad. Though I actually thinks it's nice the way it is.


OrenjiAkira:
Target selection gets a little weird with this. Maybe we need to add a 'selectable tile' texture for these states.

Also, oddly, this is a bit laggy on my computer.

PR #706, opened 2018-05-12
Enhance longwalk

### LINKED ISSUE #609: Make it possible to stop a longwalk
With any other input.

OrenjiAkira:
This PR:
+ Makes longwalk cancelable (press any input to cancel, please update external libs for this)
+ Makes longwalk cancel itself in the presence of harmful bodies
+ Makes longwalk cancel itself when surrounded by a different terrain.

It does not implement drop checkers, because as I did it I realised that drops would need to be unique (have an id, be an actual object), and I wanted to check with the rest of the team before adding more domains to the game (it would involve creating a domain for A Log Of All Seen Drops So Far).

closes #609

### DIFF
--- a/game/domain/actor.lua
+++ b/game/domain/actor.lua
@@ -408,6 +408,19 @@ function Actor:canSee(target)
   return visible and visible > 0
 end
 
+function Actor:getHostileBodies()
+  local visible_bodies = self:getVisibleBodies()
+  local hostile_bodies = {}
+  local actor_body_faction = self:getBody():getFaction()
+  for body_id in pairs(visible_bodies) do
+    local body = Util.findId(body_id)
+    if body:getFaction() ~= actor_body_faction then
+      table.insert(hostile_bodies, body)
+    end
+  end
+  return hostile_bodies
+end
+
 function Actor:purgeFov(sector)
   local fov = self:getFov(sector)
   if not fov then
--- a/game/gamestates/animation.lua
+++ b/game/gamestates/animation.lua
@@ -1,7 +1,4 @@
 
-local DB           = require 'database'
-local DIR          = require 'domain.definitions.dir'
-local DIRECTIONALS = require 'infra.dir'
 local INPUT        = require 'input'
 
 local state = {}
@@ -36,19 +33,9 @@ function state:update(dt)
 
   MAIN_TIMER:update(dt)
 
-  if not _alert then
-    for _,dir in ipairs(DIR) do
-      if DIRECTIONALS.wasDirectionTriggered(dir) then
-        _alert = true
-      end
-    end
-    for input in pairs(DB.loadSetting('controls').digital) do
-      if INPUT.wasActionPressed(input) then
-        _alert = true
-      end
-    end
+  if INPUT.wasAnyPressed(0.5) then
+    _alert = true
   end
-
 
   if not _sector_view:hasPendingVFX() then
     SWITCHER.pop(_alert)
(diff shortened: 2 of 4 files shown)


rilifon:
Nice PR. About the drops, instead of checking for all drops you have seen
so far, just check if a "new drop has appeared in your vision", maybe using
that idea of a "seen drops" tables, and when a new drop is added, it
cancels longwalk if its active.

Anyway, we cant think about this in the future

Em Sáb, 12 de mai de 2018 01:56, W. Akira Mizutani <notifications@github.com>
escreveu:

OrenjiAkira:
@kazuo256 Yeah, I played some crawl to see how the long walk works, and ours is too sensitive indeed. I will try to make it only check for terrain after walking at least once. I will also check if the failing every other time is not some variation of this.

@rilifon the thing about a 'seen drops' table is that every drop has to be unique, so as not to add them multiple times. Thus they would need ids. I got halfway through implementing it when I realised this, but maybe I didn't think it through very well.

rilifon:
But can't you use positions of drops? Their (i,j) are an unique identifier

Em Dom, 13 de mai de 2018 12:32, W. Akira Mizutani <notifications@github.com>
escreveu:

OrenjiAkira:
> Their (i,j) are an unique identifier
Yes, but only until the amount of drops in that position changes. What if a monster killed a body and it dropped something over a tile you saw earlier which had a drop in it already? What if the monster got the first drop, and then made something else drop the exact same thing in the same spot, but of a different instance?

rilifon:
Everytime a drop is added or removed you update your list, and you store
(i,j,k), where k is the amount.

Em Dom, 13 de mai de 2018 13:00, W. Akira Mizutani <notifications@github.com>
escreveu:

rilifon:
But my suggestion is a list based onmy what you see, in actor fov. In my
opinion, stopping long walk for a drop is only because this drops could be
interesting for me, it doesn't matter if I seen him before, it may well be
the case but I only need it now.

But after all those additions on what could stop longwalk, even without
playing it seems really sensitive. It could be interesting to have a
settings menu in-game which you can check which options you would like to
interrupt a long walk. This way it could be moddable and versatile to your
current needs (looking for a pack, looking for food, looking for the exit,
etc)

Em Dom, 13 de mai de 2018 13:20, W. Akira Mizutani <notifications@github.com>
escreveu:

OrenjiAkira:
I'm working on this branch rn. I will add the WIP tag so no one will merge it yet then.

OrenjiAkira:
I'm trying to figure out what DCSS's adjacency algorythm is, but I can't seem to nail it down.

OrenjiAkira:
Ok, so in this last push I made the adjacency verify only the three tiles toward which you walk, instead of all 8 adjacent tiles. This makes it quite close to crawl's. Also I refactored a little the update function of `user_turn` to make it cleaner and less prone to bugs.

OrenjiAkira:
@Kazuo256 what do you think about the drop discussion?

Kazuo256:
Regarding the drop stuff, I think stopping whenever a drop enters the FOV, even if we'd already seen it, is a valid intermediate solution until we know if we want to assign IDs to drops. However, since it would only be temporary, and this PR seems like a lot of work already, I think it's too much of a hassle for not much gain right not. I prefer to merge the PR the way it is and just discuss it all properly.

rilifon:
If everything is set, someone else please delete the branch

Issue #723, opened 2018-05-18
Redesign HUD


OrenjiAkira:
Set what info shows up where and how during gameplay.

Kazuo256:
Divided to #737

Issue #657, opened 2018-04-12
Proposal: alternative attribute system


Kazuo256:
![asd](https://cdn.discordapp.com/attachments/200674568921153536/433769134128365572/JPEG_20180411_202338.jpg)

Kazuo256:
Arbitrary adjustment: how about we make intermediate attributes (DEF, EFC, and VIT) each depend on *two* primary attributes? Like so:

```lua
DEF = (COR + ARC) / 2 -- being fit and being smart gives you more opportunities to reduce damage
EFC = (ARC + ANI) / 2 -- being ingenius and engaged allows you to make better use of resources
VIT = (ANI + COR) / 2 -- being strong-willed and strong-muscled makes you a survivor
```

Pros:
1. This naturally nerfs DEF as we wanted to, since to get high DEF you need a weaker damage output (divided between COR and ARC)
2. More interesting mixed choices of builds

Cons:
1. More confusing?
2. Nothing stops me from having more and more random ideas

Kazuo256:
In case we do the above, I'll leave some notes here:

### Aptitudes would now go from -2 to +2

### HP formula needs adjustment:
```lua
HP = 20 + (4+CON)*VIT^2 - (7+CON)*VIT
```
### Damage reduction formula is `(DEF)d(size(RES))`, where

| RES | size(RES) |
| --- | --- |
| -2 | d4 |
| -1 | d6 |
| 0 | d8 |
| 1 | d12 |
| 2 | d20 |

### WIP

PR #732, opened 2018-06-10
New layout

OrenjiAkira:
Moves a bunch of stuff of the layout.

Related to #723 

Changes:
+ Most of the actor's stuff is now displayed in a sidebar to the right
+ I refactored all of actorview, so now the "hold [Q] for more info" input does not work anymore
+ The minimap and main attributes are displayed at all times
+ Widgets are classified visually into three groups: placements, traits, and conditions; their icons is displayed in tiny squares in the sidebar (it looks good! i think)
+ Exp gains and deductions are previewed next to the exp display, now right below the minimap
+ When you use an upgrade card, the attribute bars animate until it displays how much exp is left; if you pass an attribute level, it goes BLING and the bar starts over from empty (better seen than explained)
+ Now drops do not normally give you VIT or DEF upgrade cards, although there is still a minor chance of getting one; I did this in accordance to #657, but also because it doesn't fit into the HUD

I wanted to also implement the ability belt, but this was getting too big already.

Please test! Some things might not working exactly as before, but we should fix that in future PRs.


game/draw.lua
```diff
@@ -17,7 +17,7 @@ function draw.allTables()
 
   DrawTable(DRAW_TABLE.BG)
 
-  CAM:attach() --Start tracking camera
+  CAM:attach(nil, nil, 960) --Start tracking camera
```
Kazuo256: Is there a constant or something we can use here?

game/view/actor.lua
```diff
@@ -88,186 +69,90 @@ function ActorView:draw()
...
+  -- character name
+  FONT.set("TextBold", 24)
+  g.translate(_PANEL_MG, _PANEL_MG)
+  g.setColor(COLORS.NEUTRAL)
+  g.printf(actor:getTitle(), 0, -8, _PANEL_INNERWIDTH, "left")
+  -- character hp & pp
+  g.translate(0, 48)
+  ACTOR_HEADER.drawBar(g, "HP", hp, max_hp, COLORS.SUCCESS, COLORS.NOTIFICATION)
+  g.translate(0, 32)
+  ACTOR_HEADER.drawBar(g, "PP", pp, max_pp, COLORS.PP, COLORS.PP)
```
Kazuo256: If I understood right, this method purposefully leaves the current graphics transformation translated, which in turn influences things drawn thereafter. This seems to happen in a few other places as well.

Is it possible to write it in a way that uses `push()` and `pop()` so it's easier to modify the code in the future without messing with the whole layout?

(if it's too much, we can just create an issue for this instead)

game/view/actor/attr.lua
```diff
@@ -0,0 +1,148 @@
...
+  particles:setSpeed(256)
+  particles:setSpread(2*pi)
+  particles:setColors(COLORS.NEUTRAL, COLORS.TRANSP)
+  particles:setSizes(4)
+  particles:setEmissionArea('ellipse', 0, 0, 0, false)
+  particles:setTangentialAcceleration(-512)
+  return particles
+end
+
+local function _renderAttribute(data)
```
Kazuo256: Why can't the coroutine just receive three parameters (`actor`, `attrname`, and `particles`)?

Kazuo256: Also, the `particles` is redundant, isn't it? You can get it from `actor` and `attrname`

game/view/actor/attr.lua
```diff
@@ -0,0 +1,148 @@
...
+    -- yield
+    coyield(lvl, offset, progress)
+  end
+end
+
+local ATTR = {}
+
+function ATTR.init(width)
+  _barwidth = width/4
+  _font = FONT.get("Text", 20)
```
Kazuo256: These two are essentially constant (`_barwidth` maybe not so much, but even then). so could we write them in uppercase?

game/view/actor/attr.lua
```diff
@@ -0,0 +1,148 @@
...
+  end
+  g.rectangle("fill", 0, 0, _barwidth, 16)
+  if not rise then 
+    g.setColor(COLORS[attrname])
+  end
+  g.rectangle("fill", 0, 0, percent*_barwidth, 16)
+  g.setColor(COLORS.NEUTRAL)
+  g.pop()
+  --_rot = fmod(_rot - 0.5*pi*_dt, 2*pi)
+  g.draw(particles, 48, -_font:getHeight()/2, _rot)
```
Kazuo256: The `_rot` variable doesn't seem to be used anywhere, and is always equal to zero.

game/view/actor/minimap.lua
```diff
@@ -0,0 +1,110 @@
...
+    1, 1,
+    0, 1,
+  }
+  _body = {
+    0.5, 0,
+    1, 0.5,
+    0.5, 1,
+    0, 0.5,
+  }
+  _font = FONT.get("Text", 20)
```
Kazuo256: Again, some of these behave like constants, so it's better to uppercase them.

Also, the `_tile` and `_body` variables could use a `_polygon` or `_vertices` suffix.

game/view/actor/widgets.lua
```diff
@@ -0,0 +1,107 @@
...
+local function _getConditions(body)
+  local conditions, n = {}, 0
+  for i,widget in body:eachWidget() do
+    if not widget:getWidgetPlacement() and not widget:isWidgetPermanent() then
+      n = n + 1
+      conditions[n] = widget
+    end
+  end
+  return conditions
+end
```
Kazuo256: Let's open an issue to refactor these into a single function sometime.

game/view/actor/widgets.lua
```diff
@@ -0,0 +1,107 @@
+
+local RES        = require 'resources'
+local PLACEMENTS = require 'domain.definitions.placements'
+local COLORS     = require 'domain.definitions.colors'
+local FONT       = require 'view.helpers.font'
+
+local _font
+local _mg = 24
+local _pd = 4
+local _sqrsize = 36
```
Kazuo256: Constants without uppercase

game/view/actor/widgets.lua
```diff
@@ -0,0 +1,107 @@
...
+
+local WIDGETS = {}
+
+function WIDGETS.init()
+  _font = FONT.get("Text", 20)
+  _widgetgetter = {
+    _getPlacements,
+    _getTraits,
+    _getConditions
+  }
```
Kazuo256: Uppercase

game/view/helpers/exp.lua
```diff
@@ -0,0 +1,59 @@
...
+local delta = love.timer.getDelta
+
+
+local _OFFSET_SPEED = 120
+
+local _card
+local _offset = 0
+local _darkcolor = COLORS.DARK
+local _positivecolor = COLORS.VALID
+local _negativecolor = COLORS.WARNING
```
Kazuo256: Uppercase

Kazuo256:
Please mark the PR as WIP if you intend to push new commits XD

game/view/actor/attr.lua
OrenjiAkira: there is nowhere else in the code where we use fonts as a constant. they are objects loaded and cached in runtime

game/view/actor.lua
OrenjiAkira: this way is better because each rendering section has its own relative positions, and i don't have to recalculate its base position at the beginning

game/view/actor/attr.lua
OrenjiAkira: they can change in runtime. i think i mess with the lineheight in sometimes. so i don't think uppercasing them at all is a good idea

OrenjiAkira:
I think most of the requested changes are done. The ones I have not done I either explained why, or they are just for future reference (making issue out of it)

game/view/actor/attr.lua
Kazuo256: All right, makes sense.

PR #768, opened 2018-06-22
Ability belt

### LINKED ISSUE #451: Ability belt hud?
This is a hud proposal, which is creating an "ability belt" that is navigatable via shoulder buttons (L & R in console controllers). It is basically a list of "available activatable abilities". This is a list of the signature ability + all widget abilities you can perform in a turn. This basically sets a state of the "current ability" to point towards the ability of the list.

And as such, this makes selecting a widget to activate much more user friendly. It would be somewhat like mieu ring in tales of the abyss, or item selection in zelda TP (wii version). We'd have to design some cool interface for it though

OrenjiAkira:
Now we have an ability belt to "equip" a widget activatable ability!

The default input is `ACTION_2`. which defaults to `[E]` key. Quick press is a quick change. Hold and move and you can choose whichever ability available at the time. Release or interact input will confirm the selected ability.

Closes #451 

This also implements a default trait widget for the three base classes, but not for monsters. To make available for monsters, we need to tune their AI accordingly.

The one unfortunate thing about this feature is that, specifically when you have exactly 2 abilities available, the selection alternates between going up and down in the belt. Try it to see what I mean.


game/view/readyability.lua
```diff
@@ -0,0 +1,232 @@
...
+end
+
+function ReadyAbilityView:getSelection()
+  return self.selection
+end
+
+function ReadyAbilityView:selectNext()
+  local previous = self.selection
+  self.selection = _next(self.selection, self.widget_count, 1)
+  self.offset = _dist(self.selection, previous, self.widget_count)
```
Kazuo256: If you do
```lua
-- in selectNext()
self.offset = math.abs(_dist(...))
-- in selectPrev()
self.offset = -math.abs(_dist(...))
```

Then it properly animates for two selectable abilities. It assumes you never move more than 1 at a time, which seems reasonable given how this is coded.

Kazuo256: Actually, you can just do
```lua
self.offset = 1 -- in selectNext
self.offset = -1 -- in selectPrev
```
Because the line before you explicitly move `self.selection` by either +1 or -1.

OrenjiAkira:
Ok, now 2 abilities work just fine

Kazuo256:
Don't know if you saw my second comment, but there's a simpler solution.

OrenjiAkira:
Seen and done it now

PR #771, opened 2018-06-26
Allow neutral VIT-based potion card

Kazuo256:
Also:
1. Use same range-based formula from damage in healing effects
2. Allow cards with attribute "NONE"
3. Body:heal() returns effective healed amount
4. Add some helpful attribute enums
5. Rename DEFS.ATTR.DMG -> DEFS.ATTR.POWER_RANGE
6. Swap DEFS.ATTR.POWER_RANGE parameters (and fix calling sites)
7. Add `get_body_attribute` operator
8. Make pack cards have the player as their owner even before they
   are actually added


Kazuo256:
Oh right, I also changed the meditation card to stun the player until its effects ends instead of ending on the first act. We might need to balance the amount of charges.

game/domain/definitions/attribute.lua
```diff
@@ -10,9 +10,16 @@ ATTR.INFLUENCE = {
...
 
-function ATTR.DMG(attr, base)
-  local power = attr * base
-  return power/4, 3*power/4
+local floor = math.floor
+local function round(x, ...)
+  if x then
+    return floor(x + 0.5), round(...)
+  end
+end
```
OrenjiAkira: you do know we already have a round function in `common.math` right
you can just expand that function to be recursive as well instead of reimplementing it

game/database/domains/card/calming_down.json
```diff
@@ -12,8 +12,10 @@
         "effects":[{
             "amount":5,
```
OrenjiAkira: i think 'amount' field is now deprecated yes?

game/database/domains/card/fix-turret.json
```diff
@@ -3,31 +3,38 @@
...
           "output":"turret"
+        },{
+          "type":"operator",
+          "output":"arc",
+          "name":"get_attribute",
+          "which":"ARC"
         }],
       "effects":[{
-          "type":"effect",
+          "amount":4,
```
OrenjiAkira: deprecated 'amount' field

game/database/domains/card/heal.json
```diff
@@ -3,51 +3,40 @@
...
-          "sides":"=max"
+          "which":"ANI"
         },{
           "type":"operator",
           "pos":"=body-pos",
           "name":"get_body_at",
           "output":"self"
+        }],
+      "effects":[{
+          "amount":"=amount",
```
OrenjiAkira: again here

OrenjiAkira:
Also it crashed when I tried to heal myself with a potion

PR #789, opened 2018-06-28
Fix widget duplication

Kazuo256:
When you equipped a widget over another, the previous widget was
added twice to the buffer.


OrenjiAkira:
I have not experienced this bug, and the code you remove is actually important (without it the widget just disappears!).

PR #794, opened 2018-08-04
Fix custom input methods for gamestates

OrenjiAkira:


OrenjiAkira:
I will fix that gimme a sec

PR #797, opened 2018-08-04
Fix deletion/saving when renaming specs in DB

### LINKED ISSUE #773: Erase unused fields
If you rename a field in some schema then edit something with it, when you save, the old field remains there in the json.

OrenjiAkira:
Renaming/deleting things in DB is now delegated better to DB, and uses its own methods in a smarter way.

### DIFF
--- a/game/database/init.lua
+++ b/game/database/init.lua
@@ -264,6 +264,23 @@ function DB.save(container)
   _save(container, basepath)
 end
 
+function DB.renameGroupItem(category, group_name, oldname, newname)
+  local group = _loadGroup(category, group_name)
+  local item = group[oldname]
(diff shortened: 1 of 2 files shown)


Kazuo256:
#773 referred to spec fields, as in you rename the actor schema's "arcana" to "magic", but specs keep their old "arcana" fields.

If I understood the code, this PR provides a more consistent way of removing and renaming entries in a DB collection, right? That is, you can now properly rename the specs themselves (but not their schema fields). Still valid, just maybe not related to that particular issue.

OrenjiAkira:
Alright, I misunderstood the issue. It wasn't very clear.

PR #798, opened 2018-08-05
Do not use mysterious global "actor"

Kazuo256:
Oooops.


rilifon:
Don't ask difficult questions

Em Dom, 5 de ago de 2018 16:46, W. Akira Mizutani <notifications@github.com>
escreveu:

Kazuo256:
It showed the same attributes for every body. That's what hinted something was wrong to me.

PR #867, opened 2018-08-30
Do not hide hand after using draw hand "card"

### LINKED ISSUE #750: Using the "draw hand" card should keep the hand open instead of closing it

Kazuo256:
Closes #750 

I don't like having to add a flag to both USER_TURN and SELECT_CARD, but am not sure if there's an reasonable alternative. Any ideas?


OrenjiAkira:
shit works, it's not that dirty, if there is ever a problem, we'll deal with it then

PR #873, opened 2018-08-30
Keep your hands up

Kazuo256:
Depends on #869 and #870


OrenjiAkira:
there is something wrong with the sound when you play a card: the sfx plays only when it goes back to the player's turn, very delayed

PR #898, opened 2018-09-13
Discard transmission effect

### LINKED ISSUE #896: Transmissions for discard your hand when drawing a new one

Kazuo256:
Depends on #895 

Closes #896

### DIFF
--- a/game/view/hand.lua
+++ b/game/view/hand.lua
@@ -45,6 +45,7 @@ function HandView:init(route)
   self.cardinfo = CardInfo(route)
   self.alpha = 1
   self.hiding = false
+  self.keep_focused_card = false
 
   self:reset()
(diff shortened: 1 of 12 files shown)


game/view/hand.lua
```diff
@@ -119,12 +112,33 @@ function HandView:positionForIndex(i)
...
+  self.hiding = true
+end
+
+function HandView:show()
+  self.hiding = false
+end
+
 function HandView:update(dt)
-  self:reset()
+  local _FADE_SPD = 2
```
OrenjiAkira: why is this variable declared like a file constant?

Kazuo256: I think we're going to move it up but forgot. On it.

PR #900, opened 2018-09-13
Gameover animation state

### LINKED ISSUE #107: Death screen
Make a proper death screen, that leads you to the menu after some input

OrenjiAkira:
close #107


game/gamestates/gameover.lua
```diff
@@ -0,0 +1,26 @@
+
+local PlayerDeathView = require 'view.playerdeath'
+
+local state = {}
+
+function state:enter(from, player)
+  self.view = PlayerDeathView(player)
+  self.view:addElement("HUD")
```
Kazuo256: I think state functions are not called as methods, and `state.enter()` actually receives the previous state as its first parameter, so this code does something weird that ends up working in the end.

OrenjiAkira: [https://hump.readthedocs.io/en/latest/gamestate.html](nope)

Kazuo256: Oh yes,  you are right. I got confused because they are used like objects, but we create them like modules. Also, why dat url.

PR #945, opened 2018-10-18
Don't show holdbar when out of pp

### LINKED ISSUE #944: Disable hold bar for drawing new hand when player doesn't have enough pp

ArthurTui:
Side effect: can't buy consecutive hands by holding down the button

Closes #944

### DIFF
--- a/game/view/action_hud.lua
+++ b/game/view/action_hud.lua
@@ -71,7 +71,7 @@ function ActionHUD:init(route)
   self.holdbar:lock()
   self.holdbar:addElement("HUD")
   self.justheld = false
-  self.dont_unlock_holdbar = false
+  self.unlock_holdbar = true
 
   -- Long walk variables
   self.alert = false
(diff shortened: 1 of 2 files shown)


game/view/action_hud.lua
```diff
@@ -71,6 +71,7 @@ function ActionHUD:init(route)
   self.holdbar:lock()
   self.holdbar:addElement("HUD")
   self.justheld = false
+  self.dont_unlock_holdbar = false
```
OrenjiAkira: I don't like this double negative.

game/view/action_hud.lua
```diff
@@ -338,12 +339,21 @@ end
 
 local function _endInspect(self)
   self.inspecting = false
-  if not self.justheld and self.holdbar:isLocked() then
+  if not self.justheld and self.holdbar:isLocked() and not self.dont_unlock_holdbar then
```
OrenjiAkira: Actual triple negative, huh

ArthurTui:
I liked your third suggestion better, since the flag doesn't actually represent the holdbar's status, just wheter or not it should be locked.

PR #957, opened 2018-11-22
Sound/polyphony

### LINKED ISSUE #929: Polyphony support for SFXs
The same SFX can only have one playback at a time.

rilifon:
Closes #929

### DIFF
--- a/game/database/schema/sfx.lua
+++ b/game/database/schema/sfx.lua
@@ -8,5 +8,6 @@ for _,file in ipairs(filelist) do
 end
 
 return {
-  { id = 'filename', name = "Filename", type = 'enum', options = files }
+  { id = 'filename', name = "Filename", type = 'enum', options = files },
+  { id = 'polyphony', name = "Polyphony", type = 'integer', range = {1} }
 }
--- a/game/resources/sfx.lua
+++ b/game/resources/sfx.lua
@@ -29,6 +29,12 @@ function Source:setVolume(vol)
   end
 end
 
+function Source:seek(vol)
+  for i = 1, self.buffer_size do
+    self.buffer[i]:setVolume(vol)
+  end
+end
+
 function Source:getVolume()
     return self.buffer[1]:getVolume()
 end
--- a/game/resources/sfx.lua
+++ b/game/resources/sfx.lua
@@ -17,8 +17,6 @@ function Source:init(path, polyphony)
 end
 
 function Source:play()
-<<<<<<< HEAD
-<<<<<<< HEAD
     local source = self.buffer[self.next_index]
     source:play()
     self.next_index = (self.next_index % self.buffer_size) + 1
@@ -26,24 +24,6 @@ function Source:play()
 end
 
 function Source:setVolume(vol)
-  for i = 1, self.buffer_size do
-    self.buffer[i]:setVolume(vol)
-  end
-=======
-    self.buffer[self.next_index]:play()
-    self.last_index = self.next_index
-    self.next_index = (self.next_index % self.buffer_size) + 1
-    return self.last_index
->>>>>>> Some improvements to code
-=======
-    local source = self.buffer[self.next_index]
-    source:play()
-    self.next_index = (self.next_index % self.buffer_size) + 1
-    return source
->>>>>>> Removed unnecessary last_index variable
-end
-
-function Source:seek(vol)
   for i = 1, self.buffer_size do
     self.buffer[i]:setVolume(vol)
   end
--- a/game/resources/sfx.lua
+++ b/game/resources/sfx.lua
@@ -14,10 +14,10 @@ function Source:init(path, polyphony)
     end
 
     self.next_index = 1
-    self.last_index = self.buffer_size
 end
 
 function Source:play()
+<<<<<<< HEAD
 <<<<<<< HEAD
     local source = self.buffer[self.next_index]
     source:play()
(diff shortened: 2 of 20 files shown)


game/resources/sfx.lua
```diff
@@ -0,0 +1,49 @@
+local Sfx = {}
+
+local Source = Class{}
```
OrenjiAkira: Please use parenthesis and avoid lua's sugar function calls

game/resources/sfx.lua
```diff
@@ -0,0 +1,49 @@
...
+    for i = 1, self.buffer_size do
+      self.buffer[i] = love.audio.newSource(path, "static")
+    end
+
+    self.next_index = 1
+end
+
+function Source:play()
+    self.buffer[self.next_index]:play()
+    self.next_index = (self.next_index % self.buffer_size) + 1
```
OrenjiAkira: I think this should be cached in the object

game/resources/sfx.lua
```diff
@@ -0,0 +1,49 @@
...
+
+function Source:getVolume()
+    return self.buffer[1]:getVolume()
+end
+
+---------------------------------
+
+function Sfx.new(path, polyphony)
+  return Source(path, polyphony)
+end
```
OrenjiAkira: Why the wrapper? Is this a new convention?

game/database/schema/sfx.lua
```diff
@@ -8,5 +8,6 @@ for _,file in ipairs(filelist) do
 end
 
 return {
-  { id = 'filename', name = "Filename", type = 'enum', options = files }
+  { id = 'filename', name = "Filename", type = 'enum', options = files },
+  { id = 'polyphony', name = "Polyphony", type = 'integer', range = {1} }
```
OrenjiAkira: I don't think this should be a part of the schema.

game/resources/sfx.lua
rilifon: but we use it all the time! Look at our requires for instance

game/database/schema/sfx.lua
Kazuo256: Where would it go then?

game/resources/sfx.lua
rilifon: I just left room in case we wanted other functions for this module, instead of just returning the Source class (I was initially going to return the Class directly, but other resource files such as sprite.lua have other functions, so that's my reasoning in "wrapping" haha)

game/resources/sfx.lua
rilifon: I'm sorry, I didn't understand the comment haha can you explain it with other words?

game/database/schema/sfx.lua
rilifon: Where else is a proper place to put this then?

OrenjiAkira: Shouldn't this be a constant? I find it strange that we want such a fine tuned cache for each sfx.

game/resources/sfx.lua
OrenjiAkira: Ok but we _only_ use it for requires

game/resources/sfx.lua
OrenjiAkira: We should store the current index/last index instead of recalculating all the time. It makes things clearer.

game/resources/sfx.lua
OrenjiAkira: I'm sorry but I didn't understand that. Could you try phrasing it differently?

game/resources/sfx.lua
rilifon: Okay

game/resources/sfx.lua
rilifon: We are storing the current index, but okay, we can also store the last index. But this won't change this line you are commenting, after playing a sfx we will have to update them, and that is the only time we have to update them I think.

game/resources/sfx.lua
rilifon: In case we want to add other auxiliary functions to this module, I'm not returning the Source class directly. That's why there is a function to create a Source and return it.

game/database/schema/sfx.lua
Kazuo256: We could move it to a global config variable then, but it would still be in the schema of something, because data-driving it should be better than hard-coding it.

But, from a sound design perspective, you might actually want to enforce a limit to the polyphony of certain sounds if, for instance, they are too annoying or do not overlap well. To support this argument, all audio middleware systems I've studied before allow you set a per-sample polyphony, so it's probably something they consider worthwhile.

game/resources/sfx.lua
OrenjiAkira: Yeah I'm sorry I think I highlighted the wrong line. What I meant is that we do both the 'get next index' operation AND the 'get last index' operation. Ideally we want only to do one of them, if we can avoid it

game/resources/sfx.lua
OrenjiAkira: oh ok

game/database/schema/sfx.lua
OrenjiAkira: Alright, then it's ok the way it is

Kazuo256:
@OrenjiAkira I know you are busy right now, but can you see if there's anything else you want changed here?

The lighter cache thing is probably worth a issue of its own.

OrenjiAkira:
just fix the conflicts and merge away

PR #996, opened 2019-03-27
[PROPOSAL] Change DEF mechanic

Kazuo256:
I wasn't liking how the current DEF mechanic only boosts gained armor, since it only happens when you actually gain armor.

This PR proposes a BLOCK mechanic, which uses DEF to introduce a chance of halving incoming damage. It also moves body attributes around, such that

+ HP depends on VIT and RES (instead of CON)
+ STAMINA depends on EFC and CON (instead of FIN)
+ BLOCKCHANCE (old ARMORBONUS) depends on DEF and FIN (instead of RES)


rilifon:
Seem reasonable. Haven't played enough with defense build to notice how this change the gameflow for this chars. @OrenjiAkira can you playtest and tell us what you think?

game/domain/body.lua
```diff
@@ -397,22 +397,18 @@ function Body:removeAllArmor()
...
-  local blocked = math.min(amount, self.armor)
-  self.armor = self.armor - blocked -- should never go negative
-  local dmg = math.max(0, amount - blocked)
+  local blocked = false
+  if RANDOM.generate(100) <= self:getBlockChance() then
+    amount = math.floor(amount/2)
+    blocked = true
+  end
+  local absorbed = math.min(amount, self.armor)
+  self.armor = self.armor - absorbed -- should never go negative
```
OrenjiAkira: This comment begs for unit testing

game/domain/body.lua
```diff
@@ -397,22 +397,18 @@ function Body:removeAllArmor()
...
-  -- print damage formula info (uncomment for debugging)
-  --[[
-  local str = "%s is being attacked with %d damage!\n"
-              .. "> %s rolls %dd%d for %d defense points!\n"
-              .. "> %s takes %d in damage!\n"
-  local name = self:getSpec('name')
-  print(str:format(name, amount, name, self:getDEF(), self:getBaseDEF(),
-                   defroll, name, dmg))
-  --]]--
+  return dmg, blocked
```
OrenjiAkira: whenever possible, avoid returning multiple values. althought it's less efficient, returning a table with a field "damage" and a field "wasBlocked" is way more readable and maintainable

game/domain/body.lua
OrenjiAkira: (if you're going to add testing, don't do it in this PR, do it in another as it's already big enough of a PR)

Kazuo256: Yes, this line was actually already in the code, so it's really unrelated.

Kazuo256: Created issue #1001 for this

game/domain/body.lua
Kazuo256: Done

PR #1000, opened 2019-03-27
Talking NPCs

rilifon:
Added dialogue for bodies, allowing them to say the darnest things!

Also, you can add several effects to text, such as:
- changing its color,
- changing its size
- adding effects like shake or wave
- add pauses
- and not much more!
Documentation on available tags here: https://github.com/uspgamedev/backdoor/blob/824d08608a0615d6b723ab45d5a3642c0c92d10b/game/view/dialoguebox.lua#L63-L107
Change lines for the dummy npcs in the starting area here: https://github.com/uspgamedev/backdoor/blob/824d08608a0615d6b723ab45d5a3642c0c92d10b/game/domain/builders/sectors.lua#L40-L43


Kazuo256:
You have conflicts ;)

rilifon:
> You have conflicts ;)

But is it approved, or do you want me to change something before merging it?

game/domain/body.lua
```diff
@@ -440,4 +443,14 @@ function Body:getKiller()
...
+--Dialogue methods
+
+function Body:getDialogue()
+  return self.dialogue
+end
+
+function Body:setDialogue(dialogue)
+  self.dialogue = dialogue
+end
+
```
OrenjiAkira: While I understand why it would be in the Body class, I'm thinking that this class is becoming bloated because of it.

See also: actor class.

See also: https://refactoring.guru/smells/large-class

game/domain/builders/sectors.lua
```diff
@@ -31,6 +31,18 @@ function BUILDER.build(idgenerator, player_data)
...
+  local npcs = {
+    BODY_BUILDER.buildState(idgenerator, "corgi", 11, 11),
+    BODY_BUILDER.buildState(idgenerator, "slime", 9, 13),
+  }
+  npcs[1].dialogue = [[
+  [speed=fast][size=big][style=shake]HELLO[pause=1.5][endl=1][style=none][speed=regular][size=regular]How are you today, [color=red]my friend?[color=regular][pause=1.5][endl=1][speed=slow]...[speed=regular]well I'm feeling totally [style=wave][color=blue][speed=medium]groooooovy[style=none]!
+  ]]
+  npcs[2].dialogue = "howdy, i like pears.[pause=1][endl=1][size=small][speed=fast]dont tell anyone my  [opacity=semi][color=red][speed=medium][style=shake]dark secret"
+
+
```
OrenjiAkira: FFS please put these strings somehwere else

game/view/dialoguebox.lua
```diff
@@ -0,0 +1,388 @@
...
+  if not type or not value then
+    error("tag didn't match [type=value] pattern")
+  end
+
+  --Remove tag from text
+  text = text:sub(0, tag_start_pos - 1) .. text:sub(tag_end_pos + 1, -1)
+
+
+  return text, type, value
+end
```
OrenjiAkira: I recommend separating the parser from the interpreter here. It seems this DialogueBox class does both.

game/view/sector.lua
```diff
@@ -211,6 +214,37 @@ function SectorView:getBodySprite(body)
   return body_sprite
 end
 
+
+function SectorView:setBodySprite(body, draw)
+  self.body_sprites[body:getId()] = draw
```
OrenjiAkira: WTF is draw? Please give it a better name

game/view/sector.lua
```diff
@@ -211,6 +214,37 @@ function SectorView:getBodySprite(body)
...
+function SectorView:getBodyDialogue(body, i, j, pi, pj)
+  local id = body:getId()
+
+  --Get appropriate position for dialogue box
+  local side
+  if pj <= j then
+    side = "right"
+  else
+    side = "left"
+  end
```
OrenjiAkira: WTF is pi and pj? Please, prefer being verbose than unclear

game/view/sector.lua
```diff
@@ -211,6 +214,37 @@ function SectorView:getBodySprite(body)
...
+    dialogue_box:setSide(side)
+  end
+  return dialogue_box
+end
+
+function SectorView:resetBodyDialogue(body)
+  local id = body:getId()
+  self.body_dialogues[id] = nil
+end
+
```
OrenjiAkira: This is the kind of stuff that smells like it should be somewhere else, and not here. I know SectorView is a handful, but bloating it even more is contributing to the problem. Avoid writing more functionality into it.

game/view/dialoguebox.lua
OrenjiAkira: Also avoid multiple returns. Prefer to return a table of data instead. It's more innefficient, but at the cost of maintainability

game/domain/body.lua
rilifon: I agree. But this PR is not meant to refactor Body, I'll leave it to another PR.

game/view/sector.lua
rilifon: I will change it to player_i and player_j

game/view/sector.lua
rilifon: As discussed in the meeting, we are leaving dialogue draw on sector, and will try to reduce it in another PR.

game/view/sector.lua
rilifon: Don't know when I created this or for what reason since no one calls this. Removing it.

game/domain/builders/sectors.lua
rilifon: Its just test text to showcase what the box can do. I'll change for a smaller text, but I won't put it somewhere else until I start working on the actual initial map

game/view/dialoguebox.lua
rilifon: Its just a parser, as the name implies. It only checks if the given value is a valid one. The arguments returned are interpreted elsewhere.

rilifon: I'll change multiple returns to be a single table.

rilifon: The logic will be harder to parse and interpretate seperatly, the way the characters in the dialogue box are being stores and used. I'll leave the function as it is for now.

PR #1016, opened 2019-04-21
Fix #1015

### LINKED ISSUE #1015: [CRITICAL] Game is crashing when you go down the first stairs

Kazuo256:


game/view/sector.lua
```diff
@@ -347,10 +347,15 @@ function SectorView:draw()
           local player = self.route.getControlledActor():getBody()
           local player_i, player_j = player:getPos()
           local body_i, body_j = body:getPos()
-          if body ~= player and TILE.dist(player_i, player_j, body_i, body_j) <= 1 then
-            table.insert(dialogue_boxes, self:getBodyDialogue(body, body_i, body_j, player_i, player_j))
-          else
-            self:resetBodyDialogue(body)
+          if body:getDialogue() then
+            if body ~= player and TILE.dist(player_i, player_j,
+                                            body_i, body_j) <= 1 then
```
OrenjiAkira: Get the distance in a previous line and then check its value on the if clause, for readability. This isn't C++.

game/view/sector.lua
```diff
@@ -347,10 +347,15 @@ function SectorView:draw()
...
+              table.insert(dialogue_boxes,
+                           self:getBodyDialogue(body, body_i, body_j,
+                                                      player_i, player_j))
```
OrenjiAkira: Same here

game/view/sector.lua
Kazuo256: That would be a separate issue, this PR just adds the `if` around the already existing code.

The part about C++ didn't add much to the discussion I think.

Issue #1017, opened 2019-04-24
Reduce code line length by using auxiliary variable


Kazuo256:
In `view/sector.lua`, [around line 530](https://github.com/uspgamedev/backdoor/blob/dev/game/view/sector.lua#L350)

See #1016 for more information.

OrenjiAkira:
Code quality should not be a separate issue if one can help it. It should be done with every single PR, otherwise PRs are useless. You could just write dumb ridiculous code and merge it and open infinite issues for "fixing it later". That's not good practice, and it quickly becomes a snowball

OrenjiAkira:
We have discussed this matter personally and agreed that this was done as an exception, although we still see refactoring practices somewhat differently.

rilifon:
Closed by #1019

PR #1019, opened 2019-04-25
Minor code enhancements

OrenjiAkira:
Fixes # #1017

Also, y'all need to start coding using a linter. There are shittons of bad code around: unused variables, shadowed values, trailing whitespaces... all kinds of bloating, deprecated, and messy stuff!

And then there's the issue of using snake_case with only 80 char columns (which is very much not respected at all in several places). snake_case goes better with larger columns, as the underscore char occupies lots of useless space. I recommend changing the standard to 100 or even 120 char columns.

### DIFF
--- a/game/gamestates/animation.lua
+++ b/game/gamestates/animation.lua
@@ -59,8 +59,6 @@ function state:update(dt)
     if not _animation_task or _animation_task:done() then
       SWITCHER.pop(_alert)
     end
-  else
-    _view.sector:updateVFX(dt)
   end
 
 end
--- a/game/domain/body.lua
+++ b/game/domain/body.lua
@@ -237,7 +237,7 @@ function Body:equip(place, card)
         break
       end
     end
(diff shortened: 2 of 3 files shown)


game/gamestates/animation.lua
```diff
@@ -59,8 +59,6 @@ function state:update(dt)
     if not _animation_task or _animation_task:done() then
       SWITCHER.pop(_alert)
     end
-  else
-    _view.sector:updateVFX(dt)
```
Kazuo256: Is it ok to remove this line? Can't test right now, but will later.

Kazuo256:
Should we open a proposal for modifications in the code standard and discuss it there?

game/gamestates/animation.lua
OrenjiAkira: yes, this function does absolutely nothing. i even removed it

PR #1027, opened 2019-05-22
Reactive MultiTracks for SoundTrack

### LINKED ISSUE #819: Create track system for background music
Calm music when nothing is happening, add more intense tracks as enemies appear or focus is active

rilifon:
Closes #819


Kazuo256:
@Makiyoshi tested the new feature and liked it, but decided that the soundtrack works better if we divide into only 2 tracks: the default track, and the danger track.

game/gamestates/settings.lua
```diff
@@ -48,8 +45,8 @@ function state:init()
   _selection = 1
 end
 
-function state:enter(from, soundtrack)
-  _soundtrack = soundtrack
+function state:enter(from)
+  _soundtrack = Util.findId("soundtrack")
```
OrenjiAkira: If it's going to be globally accessible, why not just make it a singleton?

rilifon: Done

PR #1071, opened 2019-08-05
Remove diagonals for effects range checks

### LINKED ISSUE #1060: Do not allow diagonal moves
Which includes effect ranges and visibility!

rilifon:
Depends on #1069 

Should remove all checks for diagonals in effects and ranges (using norm-1 instead of infinity-norm).

This still doesn't close #1060 since still need to change visibility ranges.

Please test


Kazuo256:
Dependency met, please rebase =)

PR #1093, opened 2019-08-19
Less equipment types

### LINKED ISSUE #1057: Less equipment types
How about we have just a wieldable and a wearable slot to make the game a little simpler?

Wieldable equipment widgets always have an activated ability, while werable never do.

rilifon:
Now we only have wieldables and wearables

Closes #1057


Kazuo256:
Can you make the icon the same dark color the "wpn", "off", etc. were before?

PR #1110, opened 2019-08-19
Refactor initial sector

rilifon:
Mainly changed text to reflect recent changes (such as removed unused controls).

Added new npc to associate card color and attribute.

Also tweaked dialogue a bit, improving where it seemed necessary.


rilifon:
I'll solve the issues.

Also, @Kazuo256, can you check the new dialogue for the new npc? I don't think I explained properly each attribute, I think you are best suited for this.

PR #1164, opened 2019-09-23
Better discard animation

rilifon:
FIxed offset problem and added some fadeout.

Also fixed a bug where card order wasn't the same between view and domain, causing some visual desyncs.

Tried to add fade-in to draw card animation, but it has some conflict with handview logic to hide cards when it wants to. Decided to leave it to the future.


Kazuo256:
The brawler deck is full of temporary cards. Can you revert them?

PR #1207, opened 2019-11-18
Actor panel stays longer when gaining exp

rilifon:


game/gamestates/play.lua
```diff
@@ -164,7 +164,7 @@ function state:update(_)
   if _route.getControlledActor() or _player then
     _view.sector:updateFov(_route.getControlledActor() or _player)
   else
-    print("oops")
```
Kazuo256: Prefer either
```lua
assert(_route.getControlledActor() or _player, "missing player")
_view.sector:updateFov(_route.getControlledActor() or _player)
```
or
```lua
if _route.getControlledActor() or _player then
  _view.sector:updateFov(_route.getControlledActor() or _player)	    
else
  return error("missing player")
end
```

rilifon: Used the second one

PR #1439, opened 2020-07-02
Added better indication for quick cards (spend half exhaustion)

### LINKED ISSUE #1429: Indicate when cards spend only half-time

rilifon:
Closes #1429


Kazuo256:
The icon gets hidden in quick condition cards (see Concentration). Also, I think it would be better with the same black shade as the rest of the card elements.

PR #1457, opened 2020-07-23
Make control hints context-aware

### LINKED ISSUE #1449: Contextual input hints
For instance, show "W: view packs" normally and "W: discard cad" when hand is raised.

Kazuo256:
Also improve layout and refactor to avoid repeated code.

Close #1449


game/view/gameplay/actionhud/controlhint.lua
```diff
@@ -0,0 +1,70 @@
+
+local FONT    = require 'view.helpers.font'
+local RES     = require 'resources'
+local PROFILE = require 'infra.profile'
+local COLORS  = require 'domain.definitions.colors'
+
+local Class   = require "steaming.extra_libs.hump.class"
+local ELEMENT = require "steaming.classes.primitives.element"
+local vec2    = require 'cpml' .vec2
```
OrenjiAkira, line 9: single quote ou double quote? don't you guys have a standard?

game/view/gameplay/actionhud/controlhint.lua
```diff
@@ -0,0 +1,70 @@
...
+
+function ControlHint:init(x, y, texture_name, hints)
+    ELEMENT.init(self)
+    self:setSubtype("control_hints")
+
+    self.pos = vec2(x, y)
+    self.hints = hints
+    self.mode = "DEFAULT"
+
+    self.show = PROFILE.getTutorial("finished_tutorial")
```
OrenjiAkira, line 23: i highly recommend to not use magical literals. always prefer to store them somewhere

game/view/gameplay/actionhud/controlhint.lua
```diff
@@ -0,0 +1,70 @@
...
+    self.alpha = 0
+    self.show_speed = 5
+    self.image = RES.loadTexture(texture_name)
+    self.image:setFilter("linear")
+    self.text_font = FONT.get("Text", 20)
+    self.text_font2 = FONT.get("Text", 16)
```
OrenjiAkira, line 29: actually this seems a recurring problem: magic values. they're bad practice, so i recommend to try not to use arbitrary values like these -- the code very easily becomes unmaintanable

OrenjiAkira: in fact, if these values are never changed, why are they even part of the instance itself, and not stored somewhere else?

game/view/gameplay/actionhud/controlhint.lua
```diff
@@ -0,0 +1,70 @@
...
+    g.setColor(1,1,1,self.alpha)
+    g.draw(self.image, x, y, nil, scale)
+
+    local gap = 4
+    local text_y = y + self.image:getHeight()*scale/2
+                     - self.text_font:getHeight()/2
+    local text_x = x + self.image:getWidth()*scale + gap
+    self.text_font:set()
+    local c = COLORS.BLACK
+    g.setColor(c[1], c[2], c[3], self.alpha)
```
OrenjiAkira, line 62: this kind of code exists in such abundance in this game that it really could be wrapped in some kind of way, don't you think?

game/view/gameplay/actionhud/hand/cardinfo.lua
```diff
@@ -31,7 +31,7 @@ function CardInfo:init(route)
   local w, h = VIEWDEFS.VIEWPORT_DIMENSIONS()
   self.route = route
   self.card = nil
-  self.position = vec2(w * 0.02, h * 0.07)
+  self.position = vec2(w * 0.02, h * 0.12)
```
OrenjiAkira, line 34: nevermind more magical values, but why is this one called `position`, while other position members in other classes are named `pos`?

game/view/gameplay/actionhud/init.lua
```diff
@@ -109,15 +105,29 @@ function ActionHUD:init(route)
   ADJACENCY.unset(self.adjacency)
 
   -- Control hints
-  self.cancel_hint = CancelHint(13, 8)
+  self.hand_hint = ControlHint(240+13, 28, "button-key-draw_hand", {
+    DEFAULT = "Open hand", FOCUS = "Close hand"
+  })
+  self.hand_hint:register("HUD_MIDDLE")
```
OrenjiAkira, line 111: HUD_MIDDLE could be a value on a table somewhere

game/view/gameplay/actionhud/init.lua
```diff
@@ -109,15 +105,29 @@ function ActionHUD:init(route)
...
+  self.open_packs_hint = ControlHint(240+138, 8, "button-key-open_packs", {
+    DEFAULT = "See packs", FOCUS = "Discard"
```
OrenjiAkira, line 113: ok i know this would require massive amounts of refactoring but... a thought: maybe visual text should be stored somewhere using unique keys (aka LocaleKeys)? this way you can organize ingame-flavour text and even expand into translation in the far future. im pretty sure this is done with the core attributes of the game in a way, isn't it?

game/view/gameplay/actionhud/init.lua
```diff
@@ -109,15 +105,29 @@ function ActionHUD:init(route)
...
-  self.open_packs_hint:register("HUD_MIDDLE")
-  self.show_stats_hint = ShowStatsHint(527, 8)
+  self.show_stats_hint = ControlHint(13, 8, "button-key-show_stats", {
+    DEFAULT = "(hold) Show stats", FOCUS = "(hold) Show stats"
+  })
   self.show_stats_hint:register("HUD_MIDDLE")
-  self.toggle_hints_hint = ToggleHintsHint(734, 8)
+  self.toggle_hints_hint = ControlHint(13, 48, "button-key-toggle_hints", {
+    DEFAULT = "Toggle hints", FOCUS = "Toggle hints"
+  })
```
OrenjiAkira, line 130: these param values seem very arbitrary to me, especially because the last one is a godamn table.
wouldn't it be better to just pass a single table for all the params? this way all params would be _named_, and we would know all the required fields of the constructor

game/view/gameplay/actionhud/init.lua
```diff
@@ -371,11 +381,12 @@ function ActionHUD:update(dt)
...
   -- Control hints
-  self.cancel_hint:update(dt)
-  self.confirm_hint:update(dt)
-  self.open_packs_hint:update(dt)
-  self.show_stats_hint:update(dt)
-  self.toggle_hints_hint:update(dt)
+  local control_hints = Util.findSubtype("control_hints")
+  if control_hints then
+    for button in pairs(control_hints) do
+      button:setMode(self.player_focused and "FOCUS" or "DEFAULT")
```
OrenjiAkira, line 387: more magical literals that would be really benefit to be stored somewhere else

game/view/gameplay/actionhud/controlhint.lua
Kazuo256: A much bigger problem than this PR alone. Opening #1458 to use luafmt as soon as possible.

game/view/gameplay/actionhud/controlhint.lua
Kazuo256: Also too big for this PR, since many other parts of the code rely on `PROFILE.getTutorial()`. Issue #1459 will address this particular case of magical literals.

game/view/gameplay/actionhud/controlhint.lua
Kazuo256: You mean, there should be constant (or, at least, load-time) font objects that this kind of code relies on?

Kazuo256: Opened issue #1460 from what I reckon you propose.

game/view/gameplay/actionhud/controlhint.lua
Kazuo256: Yes, probably. Issue #1461 addresses this.

game/view/gameplay/actionhud/hand/cardinfo.lua
Kazuo256: Because we are inconsistent. Can this be solved with luafmt? These are field names, not local variable names, so I would be surprised if it was capable of doing so.

Either way, there is now issue #1462 to handle this.

game/view/gameplay/actionhud/init.lua
Kazuo256: Added issue #1463 for this specific type of magical literal.

game/view/gameplay/actionhud/init.lua
Kazuo256: Yeah... maybe someday. (#1464)

game/view/gameplay/actionhud/init.lua
Kazuo256: That last argument is a table that maps a UI state name to what hint will be shown for that button icon. While I agree that it is by far not the best API (and that it is coupled to its internal implementation), I don't see how grouping all parameters in a single table would help in knowing which are required and which are not. They *would* be named, though, but that would also be inconsistent other constructor APIs.

Possible improvements we could do here:
+ accept a `vec2` parameter for the object position instead of a `(x, y)` "tuple" - this should make it clearer
+ maybe store icon file names in a table so the third parameters becomes something like `TOGGLE_HINT_ICON_ID`
+ alternatively, load it from the database, but that is a much larger refactoration
+ maybe use a sequence for the hints since the UI states are predetermined? that would require some constants for the indices, probably

What do you think?

game/view/gameplay/actionhud/init.lua
Kazuo256: Alright, since these literals are directly tied to this PR, I'll move them to tables before merging.

game/view/gameplay/actionhud/init.lua
OrenjiAkira: - i like the idea of storing the icon file name somewhere else. ideally it should be in the database, yeah, but just making it so it's not in a magical literal would improve this a lot.
- the vector instead of two numbers seems better too, but it does seem a bit overkill and would cause reference issues in the constructor itself
- grouping params in a single table would help because just by reading an example of usage of the API, i can easily understand what is going on, and although that in itself won't _completely_ tell me what params are required or not, they will most certainly tell me in a general mannter some or all of the params a function accepts

PR #1532, opened 2020-09-06
Persist a route at a configurable frequency

Kazuo256:
You can now set a frequency of 0% to 100%, where 100% means persisting every turn and 0% means persisting never. The game will *always* autosave upon a graceful exit though, but the route is no longer saved when the game crashes. In that case, it restores to the last turn-based autosave (or the last time you closed the game normally, if frequency is zero).

### DIFF
--- a/game/database/settings/user-preferences.json
+++ b/game/database/settings/user-preferences.json
@@ -8,5 +8,10 @@
     "default": 100,
     "range": [0, 100],
     "step": 10
+  },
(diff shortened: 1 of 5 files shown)


rilifon:
I liked the PR, so I will merge it. However, it would be nice in the future to display in the settings tab in between how many turns your game will autosave, since currently its pretty arbitrary how the user will use that setting

PR #1594, opened 2020-11-21
Change fullscreen mode

aki-cat:


Kazuo256 (approved):

Kazuo256:
This is temporary until we fix "windowed fullscreen" mode.

PR #1603, opened 2020-11-29
Fix #1596: Add git submodules

### LINKED ISSUE #1596: External dependencies that use git repositories are not version-locked
And thus any changes to the dependencies may break the game.

There are two solutions to this:

1. Save the current commit hash of dependencies while they still work
2. Switch to submodules which do just that

Alberdi:
Move all externals references to be tracked via git submodules. This way
each commit in this repository will be pointed to a particular commit of
each one of the external repositories, making it easier to ensure that
all compilations from the same commit have the same result.

When any of the external repositories change, any developer can do
"make update" and it will take the latest commit from each one. After
verifying that everything works, that same developer can push the
updates so they will be available for everyone in a new commit.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,15 @@
+[submodule "externals/STEAMING"]
+	path = externals/STEAMING
+	url = https://github.com/uspgamedev/STEAMING.git
+[submodule "externals/luxproject"]
+	path = externals/luxproject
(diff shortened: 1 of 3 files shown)


Kazuo256 (approved):
This looks excellent!

We really appreciate your contribution =)

